PR #11, opened 2015-01-09
minor clean ups

lantti:
While reading the 1.6 documentation I hit some hard to parse sentences, so I thought to try my hand at crafting a PR. Most of the mistakes I have spotted so far were already fixed in master, but at least this one was still left for me.


kxepal:
Merged, thanks! (:

PR #12, opened 2015-01-10
Add a missing function source argument to the list of arguments for the ...

lantti:
...rereduce message of the Query Server Protocol

It seems the rereduce message description text had the reduce function source argument missing. The example below was apparently still correct.

### DIFF
--- a/src/query-server/protocol.rst
+++ b/src/query-server/protocol.rst
@@ -227,13 +227,15 @@ keys and the list of values.
 ============
 
 :Command: ``rereduce``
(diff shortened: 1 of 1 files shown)


kxepal:
Merged with a bit shorter commit message, thanks!
It's good to have each line under 50/72 characters, but no longer. If you need to add some more notes about the change, the pattern should be the following:
```
<Short summary for 50/72 chars>

<Long detailed description by 72 chars per line>
```

PR #15, opened 2015-02-06
The documentation now have a style.

mar-ia:
The documentation was put together from different sources, written by different people. Then it was edited by even more people. And the result was... interesting.

The time when you turned to the documentation, seeking answers to deep questions like "How many ways are there to make a note-box?" is no more.


kxepal:
Thank you so much for fixing the mess in style for our docs! We'll keep it in clean now with the new style guide (:

PR #16, opened 2015-03-22
Adjust document ids in queries to match listings

jo:


kxepal:
Thank you!

jo:
That merge was fast :)

kxepal:
Fast and forward, actually (:

PR #20, opened 2015-06-19
Clarify provides() usage in _list function

BigBlueHat:
There should not be two sets of return semantics
for one function...but in the meantime, this
hopefully keeps people from hurting themselves...

### DIFF
--- a/src/couchapp/ddocs.rst
+++ b/src/couchapp/ddocs.rst
@@ -372,8 +372,11 @@ page:
 
 Templates and styles could obviously be used to present data in a nicer fashion,
 but this is an excellent starting point. Note that you may also use
-:func:`registerType` and :func:`provides` functions in the same way as for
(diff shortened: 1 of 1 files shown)


src/couchapp/ddocs.rst
```diff
@@ -372,8 +372,11 @@ page:
 
 Templates and styles could obviously be used to present data in a nicer fashion,
 but this is an excellent starting point. Note that you may also use
-:func:`registerType` and :func:`provides` functions in the same way as for
-:ref:`showfun`!
+:func:`registerType` and :func:`provides` functions in a simplar way as for
```
kxepal: `simplar` - ?

BigBlueHat:
s/simplar/similar

:stuck_out_tongue:

kxepal:
LGFM

PR #22, opened 2015-07-06
document _config changes

robertkowalski:
- document our new endpoint `/_nodes`
- leave notes that `/_config` returns a 404
- recommend configuration management for 2.0 if running a cluster

COUCHDB-2591 COUCHDB-2343


robertkowalski:
@janl @kxepal @rnewson :)

src/config/intro.rst
```diff
@@ -157,7 +168,7 @@ You should be careful with changing configuration via the HTTP API since it's
 easy to make CouchDB unavailable. For instance, if you'd like to change the
 :option:`httpd/bind_address` for a new one::
 
-    curl -X PUT http://localhost:5984/_config/httpd/bind_address -d '"10.10.0.128"'
+    curl -X PUT http://localhost:5984/_node/node1/_config/httpd/bind_address -d '"10.10.0.128"'
```
kxepal: node1 is not FQDN...because it miss domain name (:

src/api/server/configuration.rst
```diff
@@ -16,316 +16,14 @@
 Configuration
 =============
 
-The CouchDB Server Configuration API provide an interface to query and update
-the various configuration values within a running CouchDB instance.
+The classical CouchDB Server Configuration API located at ``/_config`` will
+return a 404 in CouchDB 2.0 and got replaced by :ref:`HTTP API <api/node>`.
 
-``/_config``
```
kxepal: Removing `/_config` is not true because it's still available on backdoor interface which is legal for usage. We shouldn't hide information about it.

src/config/intro.rst
robertkowalski: what is your suggestion to name it? `<nodename>`?

src/api/server/configuration.rst
robertkowalski: ah right

src/config/intro.rst
kxepal: `node1@example.com`, thought we need to clarify where to get these node names (reference to `/_membership`) - they aren't arbitrary FQDN's.

src/api/server/configuration.rst
janl: well, this opens a point of discussion, how do document the difference between the cluster and the node port? Do we at all?

IMHO, the focus should be on documenting the cluster port, as that’s the main public API.

We should document the node-specific API too, but I think it should be separate, or specifically marked, that it is node-port only. I’ll leave it to the people who are better at designing an information architecture for our docs, to decide what is better, my gut feeling says “separate sections”, but it can be convinced otherwise :)

kxepal: Separate sections was good solution if we had a plan to keep cluster/node interfaces for a long time. However, we have the other plan to get rid node interface at all, so creating a separate sections that will contains 80% of copy-paste is not good solution, even in short-term.

Special marks made with special sphinx directives is the simplest way to go here.

janl: oh, my thinking was that the node-only section only contains the things that are NOT on the cluster interface, like _config

kxepal: You cannot just limit yourself here with `_config`, `_compaction`, `_stats` etc. because even similar endpoints may return different responses:

`GET http://localhost:15984` :
```
{
    "couchdb": "Welcome",
    "vendor": {
        "name": "The Apache Software Foundation"
    },
    "version": "f29e590"
}
```

`GET http://localhost:15986` :
```
{
    "couchdb": "Welcome",
    "uuid": "d338f1c4b4555c9c7ef6ad5e8ec34667",
    "vendor": {
        "name": "The Apache Software Foundation"
    },
    "version": "f29e590"
}

```

`GET http://localhost:15984/_users`:
```
{
    "compact_running": false,
    "data_size": 0,
    "db_name": "_users",
    "disk_format_version": 6,
    "disk_size": 33936,
    "doc_count": 0,
    "doc_del_count": 0,
    "instance_start_time": "0",
    "other": {
        "data_size": 0
    },
    "purge_seq": 0,
    "sizes": {
        "active": 0,
        "external": 0,
        "file": 33936
    },
    "update_seq": [
        0,
        "g1AAAAFDeJzLYWBg4MhgTmHgz8tPSTV0MDQy1zMAQsMcoARTIkOS_P___7MSGXAqSVIAkkn2hFQ5gFTFE1KVAFJVT0BVHguQZGgAUkCF8wmrXABRuZ-wygMQlfcJq3wAUQlyZxYAHlFWdg"
    ]
}

```

`GET http://localhost:15986/_users`:
```
{
    "committed_update_seq": 1,
    "compact_running": false,
    "compacted_seq": 0,
    "data_size": 2196,
    "db_name": "_users",
    "disk_format_version": 6,
    "disk_size": 8376,
    "doc_count": 1,
    "doc_del_count": 0,
    "instance_start_time": "1436359348837084",
    "other": {
        "data_size": 2000
    },
    "purge_seq": 0,
    "sizes": {
        "active": 2196,
        "external": 2000,
        "file": 8376
    },
    "update_seq": 1,
    "uuid": "07df00f7940efd8e908b47c1cbb8f3f5"
}
```

janl: What I mean is that we *don’t* have to document the things that are on the node only port that have a corresponding endpoint on the cluster API. That way we only document the things that we expect people to use.

robertkowalski: I agree with @janl: we should not document the backdoor interface as we discourage people to use it, except for endpoints that we want that people use them (e.g. compaction)

Btw: is there any reason to keep the routes we don't want people to use on the backdoor ports, e.g. /{db}/_all_docs?

kxepal: API documentation is not about our position what people should use and what they shouldn't. It's about the product and what product is able to provide. Discouraging things is about best practices which could be implemented with a warning note.

Also, you forgot the case when node gets out of the cluster - then you'll not able to use cluster API to get any information from it because:

```
HTTP/1.1 500 Internal Server Error
Cache-Control: must-revalidate
Content-Length: 54
Content-Type: text/plain; charset=utf-8
Date: Sun, 12 Jul 2015 15:31:56 GMT
Server: CouchDB/bcb7c1a (Erlang OTP/18)
X-Couch-Request-ID: 3ca2e2d26b
X-CouchDB-Body-Time: 0

{"error":"nodedown","reason":"progress not possible"}
``` 

And them this happens, you need to find out how to get your data back. If we're going to hide all information about local interface and even remote it, we doom those people that will have to deal with such cases. I don't see any reason why we should make life complicated for our users while we can avoid it.

janl: > API documentation is not about our position what people should use and what they shouldn't. 

I firmly disagree with this :)

PR #24, opened 2015-08-03
note link between startkey and startkey_docid

slang800:
The fact that startkey_docid depends upon startkey isn't explicitly mentioned, and has caused confusion in issues like https://issues.apache.org/jira/browse/COUCHDB-834


rnewson:
+1

PR #25, opened 2015-08-08
Document CSRF protection

rnewson:
COUCHDB-2762

### DIFF
--- a/src/config/http.rst
+++ b/src/config/http.rst
@@ -507,6 +507,85 @@ Cross-Origin Resource Sharing
         - `COS tutorial <http://www.html5rocks.com/en/tutorials/cors/>`_
         - `XHR with CORS <http://hacks.mozilla.org/2009/07/cross-site-xmlhttprequest-with-cors/>`_
 
+Cross-site Request Forgery protection
+=====================================
+
+.. config:section:: csrf :: Cross-site Request Forgery
+
+    .. versionadded:: 2.0 added CSRF protection, see JIRA :issue:`2762`
+
+    `CSRF`, or "Cross-site Request Forgery" is a web-based exploit
+    where an attacker can cause a user agent to make an authenticated
+    form post or XHR request against a foreign site without their
+    consent. The attack works because a user agent will send any
+    cookies it has along with the request. The attacker does not see
+    the response, nor can they see the user agent's cookies. The
+    attacker hopes to gain indirectly, e.g, by posting to a password
+    reset form or cause damage by issuing a database delete request.
+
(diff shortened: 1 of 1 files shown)


src/config/http.rst
```diff
@@ -507,6 +507,71 @@ Cross-Origin Resource Sharing
...
+        }
+
+    .. config:option:: secret
+
+        All CSRF cookies are signed by the server using this value. A
+        random value will be chosen if you don't specify it, but we
+        recommend setting it yourself, especially if you are running a
+        cluster of more than one node. The secret must match on all
+        nodes in a cluster to avoid sadness.
+
```
kxepal: Would be good to have an example here as other options have.

src/config/http.rst
```diff
@@ -507,6 +507,71 @@ Cross-Origin Resource Sharing
...
+    .. config:option:: timeout
+
+        All CSRF cookies expire after `timeout` seconds. The default
+        is an hour.
+
+
```
kxepal: `make check` isn't happy because of addition newline here. Trying to keep same style all around.

kxepal:
Great addition! LGFM except examples and failed linter check.

rnewson:
the toolchain here was huge so I haven't even built it.

kxepal:
It's actually strange that it's all is required even you need just make html version. I'll check makefile rules.  But anyway, that's why travis here to help (:

rnewson:
tests pass (MacTex download continues to fail, so I've given up with local attempt, but I really wish we didn't need the 2.68GB blob of LaTeX just to generate our docs).

src/config/http.rst
```diff
@@ -507,6 +507,85 @@ Cross-Origin Resource Sharing
...
+          setRequestHeader("X-CouchDB-CSRF", "true");
+        }
+
+    .. config:option:: mandatory
+
+        CouchDB can insist on CSRF Cookie/Header for all requests
+        (except those to the welcome handler, /, so you can acquire a
+        cookie) with this setting. The default is false.
+
+            [csrf]
```
kxepal: This wouldn't render as code block, but just as a sub paragraph. Wrap it with `.. code-block:: ini`.

rnewson:
I updated it to match the other ways we do it in other sections. Emacs indicated I did something right but I'm flying blind until this 2.68gb thing arrives

kxepal:
All good. +1

rnewson:
no idea why it won't mark this as merged but it is merged.

kxepal:
Hashes seems different: 13e9832 on master vs f88582f here. Could you just close it then?

PR #26, opened 2015-08-08
Improve Makefile

kxepal:


kxepal:
cc @janl

dch:
I assume the main purpose is to able to build at least the HTML without needing PDF/latex generation, right?

+1 LGTM

kxepal:
@dch build HTML without need in TeX dependencies which are used for latex/pdf targets. And rebuild everything all the time for every changed bit when you actually don't need in. Yes (:

PR #28, opened 2015-08-31
COUCHDB-2794: Docs not fully correct for couch_httpd_auth/secret

andywenk:
rewording and added also the link to Cookie Auth section.

### DIFF
--- a/src/config/auth.rst
+++ b/src/config/auth.rst
@@ -212,7 +212,7 @@ Authentication Configuration
             [couch_httpd_auth]
             require_valid_user = false
 
-    .. config:option:: secret :: Proxy Auth secret token
+    .. config:option:: secret :: Authentication with secret token
 
         The secret token is used for :ref:`api/auth/proxy` and for :ref:`api/auth/cookie`. ::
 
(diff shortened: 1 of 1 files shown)


src/config/auth.rst
```diff
@@ -214,7 +214,7 @@ Authentication Configuration
 
     .. config:option:: secret :: Proxy Auth secret token
 
-        The secret token used for :ref:`api/auth/proxy` method. ::
+        The secret token is used for :ref:`api/auth/proxy` and for :ref:`api/auth/cookie`. ::
```
kxepal: Thanks, but look on one line above (;

andywenk: sry I don't know what you mean

kxepal: Line 215: `.. config:option:: secret :: Proxy Auth secret token`

andywenk: do you want sth like "Proxy and Cookie Auth" in line 215 ?

kxepal: I think about just generic "authentication secret token" or something, but you may figure something else and better (:

andywenk: I like that ;-)

PR #31, opened 2015-09-03
added Mango docs

KimStebel:
I converted Cloudant's "Query"/"Mango" documentation to ReST and removed (almost) all references to Cloudant from it. Please have a look and let me know if this can be included in the CouchDB docs.


kxepal:
Thanks for contribution! 

For API you need to use httpdomain to define resources and provide examples for each and probably decouple resources from usage how-to.

Also this PR contains an error:
https://travis-ci.org/apache/couchdb-documentation/builds/78597375

Not sure why travis didn't ping back here.

robertkowalski:
ping

janl:
build seems broken now, we can merge t hat after this gets fixed.

KimStebel:
Thanks for pinging, I'll see what I can do to get the issues sorted out so that we can merge this.

wohali:
Hi @KimStebel I know you may have moved on, but any interest in cleaning this up and helping us get this PR landed?

lornajane:
If @KimStebel doesn't have time, I can probably pick this up - is it just a case of trying to get the final failing checks to pass?

wohali:
Yes, we need the ASF License header in every file for sure. We also need to reconcile this with the other Mango docs that landed over here: https://github.com/apache/couchdb-documentation/pull/55 In other words, make sure the content, tone and advice match.

lornajane:
Thanks for pointing out where the other Mango docs landed, it definitely has a common history with this document!  There seems to be more detail in this version though so I think I'll try to patch over the extra pieces that make sense into the stuff @willholley already contributed, which will land as a separate PR since I can't amend this one, but I'll update this one when that happens.

wohali:
Sounds good! I don't want to lose the concept of both an API reference and a more discursive explanation of Mango, and naturally I don't want to lose the detail. Thanks for being tenacious on this one!

PR #34, opened 2015-09-15
Fix bashism in Makefile

micah:
The [[ operator is a bashism, and fails with POSIX shells (such as dash),
replace it with a simplier version that does the same thing, without the
unnecessary $(shell) nesting.


kxepal:
+1

kxepal:
Thanks!

PR #35, opened 2015-10-16
Fix for COUCHDB-2578

snowch:
COUCHDB-2578  _all_docs 'keys' parameter not documented


kxepal:
I have strange déjà vu on this fix. Thanks!

PR #38, opened 2016-01-04
To automatically generate the _id should be omitted

danielcobo:
To be consistent with text at line numbers 302-303:
- _id field is omitted in request
- an automatically generated _id is returned by CouchDB


kxepal:
Thanks! (:

PR #39, opened 2016-02-02
suppress verbose LaTeX output

basti1302:
pdflatex CouchDB.tex writes about 700 lines of log output to stdout,
which clutter the output of make when building CouchDB without adding a
lot of value. The option interaction=batchmode reduces this to 3 lines.
The full log of pdflatex is still available in
build/latex/CouchDB.log.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ latex: $(TEX)
 	$(SPHINXBUILD) -b $@ $(SPHINXOPTS) $(BUILDDIR)/$@
 
 pdf: latex $(PDFLATEX)
-	$(MAKE) -C $(BUILDDIR)/latex all-pdf
+	$(MAKE) LATEXOPTS=' -interaction=batchmode ' -C $(BUILDDIR)/latex all-pdf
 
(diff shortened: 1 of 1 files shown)


basti1302:
One thing I am not 100% sure about: This works with the TexLive distribution usually used on Linuxes. Which other Latex distributions are used on systems that use this Makefile? Do they all support this option? MacTex supports this option, AFAIK.

janl:
TexLive is fine, MacTex is based on TexLive IIRC.

janl:
e.g. this is good, +1, feel free to merge!

dch:
Generally we build the docs from git source only on a unix platform for releases, IIRC almost always a recent debian. This should mean that the other platforms only need to build from tarball.

dch:
on Windows the build process runs under cygwin and also used pdflatex although it's since been removed from cygwin due to lack of maintainers, so either we'd need to use an alternative tex distribution (miktex appears to support this flag) or just build from release tarball. either way good to merge!

basti1302:
Merged.

PR #40, opened 2016-02-02
ignore build directory instead of _build

basti1302:
`_build` was ignored but the build process actually creates a `build` directory (without leading underscore).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 ext/*.pyc
-_build/
+build/
 


dch:
+1 to merge this anyway, the failure from travis is the same as I encountered yesterday, unrelated to this straightfwd change.

kxepal:
@basti1302 you can rebase for the green build. We need to fix sphinx 1.3.5 support additionally.

basti1302:
Merged.

PR #43, opened 2016-03-16
Removing _rev

timmillwood:
Let's assume the recording of a replication checkpoint is the first time this replication has taken place, and therefore no `_rev` is needed, and the response will always be `"rev": "0-1"`.

I have also removed the content-length header, and the _revisions element, because these seem to add more confusion than explanation.

### DIFF
--- a/src/replication/protocol.rst
+++ b/src/replication/protocol.rst
@@ -1585,20 +1585,12 @@ Replicator updates Replication Log on Source:
 
         PUT /source/_local/afa899a9e59589c3d4ce5668e3218aef HTTP/1.1
         Accept: application/json
-        Content-Length: 591
         Content-Type: application/json
         Host: localhost:5984
         User-Agent: CouchDB
 
         {
             "_id": "_local/afa899a9e59589c3d4ce5668e3218aef",
-            "_rev": "0-1",
-            "_revisions": {
-                "ids": [
-                    "31f36e40158e717fbe9842e227b389df"
-                ],
-                "start": 1
-            },
             "history": [
                 {
                     "doc_write_failures": 0,
@@ -1633,7 +1625,7 @@ Replicator updates Replication Log on Source:
         {
             "id": "_local/afa899a9e59589c3d4ce5668e3218aef",
             "ok": true,
-            "rev": "0-2"
+            "rev": "0-1"
         }
 
 ...and on Target too:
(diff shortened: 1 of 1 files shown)


src/replication/protocol.rst
```diff
@@ -1585,20 +1585,12 @@ Replicator updates Replication Log on Source:
 
         PUT /source/_local/afa899a9e59589c3d4ce5668e3218aef HTTP/1.1
         Accept: application/json
-        Content-Length: 591
```
kxepal: I would prefer to explicitly show that this is not a chunked transfer. Removing content length will cause a questions.

src/replication/protocol.rst
```diff
@@ -1585,20 +1585,12 @@ Replicator updates Replication Log on Source:
...
         Content-Type: application/json
         Host: localhost:5984
         User-Agent: CouchDB
 
         {
             "_id": "_local/afa899a9e59589c3d4ce5668e3218aef",
-            "_rev": "0-1",
-            "_revisions": {
```
kxepal: Actually, `_revisions` does send to remote here. No reason to hide that.

src/replication/protocol.rst
```diff
@@ -1644,20 +1636,12 @@ Replicator updates Replication Log on Source:
...
         PUT /target/_local/afa899a9e59589c3d4ce5668e3218aef HTTP/1.1
         Accept: application/json
-        Content-Length: 591
         Content-Type: application/json
         Host: localhost:5984
         User-Agent: CouchDB
 
         {
             "_id": "_local/afa899a9e59589c3d4ce5668e3218aef",
-            "_rev": "1-31f36e40158e717fbe9842e227b389df",
```
kxepal: I would prefer to show here the case when Source and Target are out of sync of checkpoint document revisions. That's fine and could be, but doesn't really matters as like as the local document revisions. CouchDB uses `0-N`, some else compatible service may provide own ones. That's not a strict rule to follow.

timmillwood:
@kxepal Thanks for reviewing, I'll working a more thorough update and explain some of the nuances, and if source and target are out of date.

Maybe I don't understand `_revisions` but I've not seen a need to send it.

kxepal:
@timmillwood It actually does, you can check that via wireshark dump. As for the code part, `_revisions` are get injected [here](https://github.com/apache/couchdb-couch-replicator/blob/master/src/couch_replicator_api_wrap.erl#L363-L365) by `revs` option.

wohali:
@timmillwood any interest in reviving this PR? If not we should close it out.

wohali:
@timmillwood I'm going to close out this PR for now. If you'd like to revisit just submit a new PR and we'll be happy to get it into the official docs.

PR #47, opened 2016-04-06
Update performance.rst

seigel:
Need some help from people smarter than me.  I've pointed out a section of the description that doesn't make sense in the current version of the documentation.

### DIFF
--- a/src/maintenance/performance.rst
+++ b/src/maintenance/performance.rst
@@ -99,6 +99,7 @@ will be continuously replicating.
 Erlang
 ------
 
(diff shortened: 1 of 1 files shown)


src/maintenance/performance.rst
```diff
@@ -99,6 +99,7 @@ will be continuously replicating.
 Erlang
 ------
 
+[The documentation no longer speaks about maximum connections.  Not sure how to change this.]
```
kxepal: I guess by removing outdated block? (:

seigel: Did not know whether it was outdated or missing :D

PR #48, opened 2016-04-14
Fix the build, and fix the location of the docs repo in the 'Contributing' section

philip-morning:
Our documentation has been moved to its own
repository, couchdb-documentation, and some of 
the details for getting into it have changed.

For the example, I replaced Jan's username with 
my own, as using his would have been a 404, and 
using mine seemed like the next-best idea.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
   - sudo apt-get install -yqq texinfo
 
 install:
-  - pip install sphinx==1.3.4
+  - pip install sphinx
 
 script:
   - make ${TARGET}
--- a/src/api/basics.rst
+++ b/src/api/basics.rst
@@ -141,7 +141,7 @@ Request Headers
 
   The returned headers are:
 
-  .. code-block:: http
+  .. code-block:: none
 
       Server: CouchDB (Erlang/OTP)
       Date: Thu, 13 Jan 2011 13:39:34 GMT
@@ -163,7 +163,7 @@ Request Headers
 
   The headers returned include the ``application/json`` content type:
 
-  .. code-block:: http
+  .. code-block:: none
 
       Server: CouchDB (Erlang/OTP)
       Date: Thu, 13 Jan 2013 13:40:11 GMT
@@ -364,7 +364,9 @@ actually "losing" data if you don't accept that numbers are stored in doubles).
 Here's a log for a couple of the more common JSON libraries that happen to be
 on the author's machine:
 
-Ejson (CouchDB's current parser) at CouchDB sha 168a663b::
+Ejson (CouchDB's current parser) at CouchDB sha 168a663b:
+
+.. code-block:: none
 
     $ ./utils/run -i
     Erlang R14B04 (erts-5.8.5) [source] [64-bit] [smp:2:2] [rq:2]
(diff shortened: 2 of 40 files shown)


exclsr:
It appears that before committing this change, the build needs to be fixed. I will do my best to look into and resolve the situation.

I also see, now, that @eallik has already attempted a change similar to this one. Sorry for not looking at the pending pull requests. Anyway, I'll see what I can do about the build -- though I am not familiar with the system, so we'll see how far I get.

eallik:
I will be hard to get people to contribute if pull requests are simply not noticed...

robertkowalski:
sorry @eallik if we missed your commit / PR.

you can fix the failing tests (style check tests) by running `make check` locally. You can find the output of the failing job here: https://travis-ci.org/apache/couchdb-documentation/jobs/122980993

exclsr:
Thanks, @robertkowalski . I fixed the `make check` build and pushed a new commit. Looking into the pdf target now.

exclsr:
Hi friends, I fixed the build errors. 

Long story short, we need to use Sphinx >= 1.4.0, as it includes the LaTeX files we need -- newfloat.sty, which is used by, but absent from, Sphinx 1.3.4, which Sphinx fixed in 1.4. 

For your review.

exclsr:
Please let me know if there is more I can do to help out further with this pull request. I would be happy to perform more work to get something like this into the codebase, as it fixes the build.

.travis.yml
```diff
@@ -12,7 +12,7 @@ before_install:
   - sudo apt-get install -yqq texinfo
 
 install:
-  - pip install sphinx==1.3.4
```
kxepal: This was made with intention since sphinx eventually can break the build with newer versions. Let's pin some release.

src/api/basics.rst
```diff
@@ -141,7 +141,7 @@ Request Headers
 
   The returned headers are:
 
-  .. code-block:: http
+  .. code-block:: none
```
kxepal: The http language is provided by httpdomain extension. Why sphinx 1.3.5+ fails to use it? I think the proper fix need to fix the syntax highlight, not turn it off just because some bug.

kxepal:
@exclsr 
Sorry for long time waiting! I like changes about upgrading sphinx support and fixing latex, but definitely this shouldn't come with disable highlighting. There should be other, better fix that doesn't requires such high price.

exclsr:
@kxepal 
Thank you for taking a look at this! Yes, I agree that there is a decent amount of work to do, here, in order to make the documentation excellent. 

In this change set, I tried to keep highlighting whenever it was easy to do so. There are a bunch of situations where the current version of the docs specify highlighting, but it is broken. 

For example, in basics.rst above, 'http' is changed to 'none' because the two lines that follow are not valid http according to the parser, not that the httpdomain extension is missing. I am not suggesting that my change set is perfect, but I do think it is a step forward and at least calls out where the syntax highlighting was broken all along.

I suggest pinning sphinx to something after 1.4.

exclsr:
(Most of the changes that set the highlighter to 'none' are because the default highlighter for many documents is JavaScript, and the docs have code / examples that are not in JavaScript, and the parser complains.)

kxepal:
@exclsr 
Hm, I should read more carefully. Seems like not all the highlight have been disabled, but only part of it. Can only important http examples (basically API usage) be fixed? Wonder how else remained untouched without issues. For else blocks like shell scripts it's fine to have none.

Agree on pinning to 1.4.

exclsr:
Sounds good. I will fix up the API usage examples this week.

kxepal:
Thank you! (:

philip-morning:
Yes. :-) Thank you for working on CouchDB.

PR #50, opened 2016-05-25
Ignore Python3 __pycache__ dir

katsel:
A `__pycache__` directory appeared after making the docs with a Python3 version of Sphinx. Python3 stores cache files in that directory instead of *.pyc files. Should be added to .gitignore.


kxepal:
Thank you!

PR #53, opened 2016-07-22
COUCHDB-3068 - Fixes to make docs build with Sphinx 1.4

katsel:
* Remove JSON as default highlight language from conf.py, so Sphinx does not assume all non-marked code blocks are JSON per default.
* Use valid HTTP
  * add ` HTTP/1.1` to each request line
  * Pygments' http lexer currently does not recognize the COPY method, so disable highlighting for blocks that use COPY.
* Minor changes with shell code-blocks - shell-sessions and shell-scripts require different lexers.

### DIFF
--- a/src/api/document/common.rst
+++ b/src/api/document/common.rst
@@ -359,7 +359,7 @@
 
     **Request**:
 
-    .. code-block:: http
+    .. code-block:: none
 
         COPY /recipes/SpaghettiWithMeatballs HTTP/1.1
         Accept: application/json
@@ -1131,7 +1131,7 @@ or :header:`If-Match`:
 
 **Request**:
 
-.. code-block:: http
+.. code-block:: none
 
     COPY /recipes/SpaghettiWithMeatballs HTTP/1.1
     Accept: application/json
@@ -1167,7 +1167,7 @@ for the target document by appending the ``rev`` parameter to the
 
 **Request**:
 
-.. code-block:: http
+.. code-block:: none
 
     COPY /recipes/SpaghettiWithMeatballs?rev=8-6f5ad8db0f34af24a6e0984cd1a6cfb9 HTTP/1.1
     Accept: application/json
--- a/src/api/document/common.rst
+++ b/src/api/document/common.rst
@@ -359,7 +359,7 @@
 
     **Request**:
 
-    .. code-block:: none
+    .. code-block:: http
 
         COPY /recipes/SpaghettiWithMeatballs HTTP/1.1
         Accept: application/json
@@ -1131,7 +1131,7 @@ or :header:`If-Match`:
 
 **Request**:
 
-.. code-block:: none
+.. code-block:: http
 
     COPY /recipes/SpaghettiWithMeatballs HTTP/1.1
     Accept: application/json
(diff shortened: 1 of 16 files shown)


src/config/couchdb.rst
```diff
@@ -179,10 +179,13 @@ Base CouchDB Options
...
         * ``false``: The node responds normally, /_up returns a 200 response.
 
+        .. highlight:: none
+
         It is expected that the administrator has configured a load balancer
         in front of the CouchDB nodes in the cluster. This load balancer should
         use the /_up endpoint to determine whether or not to send HTTP requests
-        to any particular node. For HAProxy, the following config is appropriate ::
+        to any particular node. For HAProxy, the following config is
+        appropriate::
```
kxepal: I think you better replace `::` with `.. code-block:: none` definition. That's more error-prone solution. Same for the rest similar cases.

src/api/document/common.rst
```diff
@@ -1167,7 +1167,7 @@ for the target document by appending the ``rev`` parameter to the
 
 **Request**:
 
-.. code-block:: http
+.. code-block:: none
```
kxepal: Hm...why http lang doesn't likes COPY examples?

kxepal: We do have it supported in [local lexer](https://github.com/apache/couchdb-documentation/blob/master/ext/httpdomain.py#L629). May be worth to switch onto in instead of doing [fallback](https://github.com/apache/couchdb-documentation/blob/master/ext/httpdomain.py#L653-L656)?

kxepal:
Hm...PDF is broken. Seems like not your fault.

katsel:
@kxepal Hi! Thanks for your feedback.

1) Yup, using `code-block`  instead of :: makes sense in these cases. I will change that and do a new commit.

2) COPY is not part of standard HTTP, so maybe that's why (here's the [relevant pygments source](https://bitbucket.org/birkenfeld/pygments-main/src/e79a7126551c39d5f8c1b83a79c14e86992155a4/pygments/lexers/textfmts.py?at=default#textfmts.py-175)). One could file an issue at pygments for that, but that's not a quick solution.
Using the local lexer for http seems like a possible workaround, but how would you do that in an elegant way? Is it worth the effort just because of 3 examples that use COPY?

kxepal:
@katsel 
> Using the local lexer for http seems like a possible workaround, but how would you do that in an elegant way? 

Just use local one all the time? 

> Is it worth the effort just because of 3 examples that use COPY?

It seems we can cover them without problems and doing any coding work, I think we should try (:

katsel:
@kxepal 
OK, so the workaround is to remove [the try clause](https://github.com/apache/couchdb-documentation/blob/master/ext/httpdomain.py#L653-L655) from `httpdomain.py` code so the local lexer becomes the default? :)

Should I include that in the PR?

kxepal:
@katsel Please do if this helps. Otherwise we'll just keep your initial changes.

katsel:
Done. :)

> Hm...PDF is broken. Seems like not your fault.

What's broken about PDF?

I tried `make pdf` and it works for me. (TeX Live 2015)
The PDF looks good at first sight, it has some nice new things (such as automatic line breaks for long lines of code), but I haven't looked at it in detail.

kxepal:
@katsel 

> Done. :)

Awesome! Thank you, +1

> What's broken about PDF?
> I tried make pdf and it works for me. (TeX Live 2015)

Same story for me. It seems like some issue about packages on Travis. I worked before, but not today. Let's handle this in separate iteration.

katsel:
@kxepal: I rebased my commits and unpinned the sphinx version. Building the PDF works now.

kxepal:
@katsel awesome! thank you!

PR #61, opened 2016-09-07
remove references to temp views

willholley:
possibly this is too brutal but, since they are not supported in 2.0, here's a first pass at updating the docs.


rnewson:
+1

PR #63, opened 2016-09-09
Remove row-spanning in _find documentation

willholley:
Man pages do not support table row spanning, so don't do it.


wohali:
Assuming the Travis CI build passes, +1

wohali:
Ta very much :)

PR #67, opened 2016-09-21
add _cluster_setup documentation

garrensmith:


janl:
you can leave out the initial PW setup, enable_cluster will do that for you

otherwise +1

PR #88, opened 2016-11-15
Fix os_process_limit documentation

jaydoane:
Update documentation with correct description of os_process_limit
parameter function, correct default value, and hints for typical
production values.

COUCHDB-3238

### DIFF
--- a/src/config/query-servers.rst
+++ b/src/config/query-servers.rst
@@ -82,18 +82,19 @@ Query Servers Configuration
             [query_server_config]
             commit_freq = 5
 
-    .. config:option:: os_process_limit :: Query Server process limit
+    .. config:option:: os_process_limit :: Query Server process hard
+                       limit
 
         Hard limit on the number of OS processes usable by Query
-        Servers. The default value is ``25``::
+        Servers. The default value is ``100``::
 
             [query_server_config]
-            os_process_limit = 25
+            os_process_limit = 100
 
         Setting `os_process_limit` too low can result in starvation of
         Query Servers, and manifest in `os_process_timeout` errors,
         while setting it too high can potentially use too many system
-        resources. Production settings are typically 10-100 times the
+        resources. Production settings are typically 10-20 times the
         default value.
 
     .. config:option:: reduce_limit :: Reduce limit control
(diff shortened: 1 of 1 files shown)


src/config/query-servers.rst
```diff
@@ -82,16 +82,19 @@ Query Servers Configuration
...
+    .. config:option:: os_process_limit :: Query Server process limit
 
-        Amount of time in seconds that the Query Server may process CouchDB
-        command::
+        Hard limit on the number of OS processes usable by Query
+        Servers. The default value is ``25``::
 
             [query_server_config]
-            os_process_limit = 10
+            os_process_limit = 25
```
rnewson: the default is 100.

src/config/query-servers.rst
```diff
@@ -82,16 +82,19 @@ Query Servers Configuration
...
 
-        CouchDB will raise `os_process_timeout` error and kill the process in
-        case the Query Server doesn't return any result within this limit.
+        Setting `os_process_limit` too low can result in starvation of
+        Query Servers, and manifest in `os_process_timeout` errors,
+        while setting it too high can potentially use too many system
+        resources. Production settings are typically 10-100 times the
+        default value.
```
rnewson: could you add a section for os_process_soft_limit too? (also defaults to 100)

jaydoane:
While the internal default is indeed 100, it's set to 25 in rel/overlay/etc/default.ini -- so I guess it depends on what you mean by "default". I'm happy to change it to 100, though.

Also, added a section on os_process_soft_limit, which could probably use better word-smithing.

rnewson:
heh, so it is, how silly. a hard limit of 25 and a soft limit of 100 (since it's not specified in default.ini), that's a nonsense setup.

jaydoane:
@rnewson thanks for the review. If you're ok with this, can you +1 and merge for me?

Also, shall I open a PR to just remove the explicit hard limit of 25 in default.ini?

rnewson:
I'll merge this. please open a PR to fix the 25 -> 100 thing in default.ini.

rnewson:
I took care of it. thanks for the doc patch.

PR #94, opened 2016-12-08
install/index.rst: More details on Single-Node-Setup

dmunch:
Stress the fact that the “Single-Node-Setup” creates the system databases for you. Add how single node setup can be done without using the wizard.


michellephung:
hi @dmunch can you pull in the latest and rebase? we updated the travis tests and now this pr should pass

dmunch:
@michellephung Done!

michellephung:
thanks!

PR #96, opened 2017-01-04
Inline comments: Improve phrasing

padyx:
The wording of the paragraph about inline comments is not optimal.
I suggest rephrasing it.

### DIFF
--- a/src/config/intro.rst
+++ b/src/config/intro.rst
@@ -141,8 +141,8 @@ improve configuration readability.
     This could be read as: "remove the `_all_dbs` parameter from the
     `httpd_global_handlers` section if it was ever set before".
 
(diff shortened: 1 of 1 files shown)


michellephung:
+1 but needs to pass the travis tests

padyx:
@michellephung Looks unrelated - the used versions of Python seem to be too old
`Error: Sphinx requires at least Python 2.7 or 3.4 to run.`
https://travis-ci.org/apache/couchdb-documentation/jobs/188771541#L399
The python versions being used: 3.3 / 2.6

I can open another PR and try to simply update the python versions?

michellephung:
@padyx if you have time to open another PR to fix the python version that would be great! :D !! Thanks in advance! If you don't I can take a look this weekend

PR #97, opened 2017-01-10
Update python versions to 2.7/3.4

padyx:
According to the latest failed builds, sphinx needs at least Python 2.7 or 3.4 to build.

This is simply an attempt to apply the solution suggested by sphinx. I don't know if there is any other impact to this repository - please review :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 python:
-  - 2.6
-  - 3.3
+  - 2.7
+  - 3.4
 
 before_install:
   - sudo apt-get update -qq
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ MAKEINFO     := makeinfo
 
 BUILDDIR     := build
 SOURCE       := src/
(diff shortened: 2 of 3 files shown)


padyx:
@michellephung Looks like this fixed the sphinx error about the wrong python version. But there are now subsequent errors:

```
Warning, treated as error:
WARNING: latex_paper_size conflicts with latex_elements['papersize'], ignored.
```
If you know how to fix that, I'll try to incorporate it into the PR.

kxepal:
@padyx 
[latex_paper_size is deprecated](http://www.sphinx-doc.org/en/1.5.1/config.html#confval-latex_paper_size), so be brave to it rid (:

kxepal:
Since it comes from Makefile, just replace it with `latex_elements.papersize` - this should works.

padyx:
Looks likes it works! Thanks for the input @kxepal .
Can anyone with a latex environment verify that the paper format is correct? Likely only relevant for PDF output which I don't get from Travis.

kxepal:
@padyx 
It's fine. Thanks you!

PR #55, opened 2016-09-01
API documentation for _find and _index endpoints

willholley:
API reference documentation for the new _find and _index endpoints.

Credit to this primarily goes to @KimStebel who wrote the original content at [https://github.com/cloudant-labs/slate].

This is heavily based on the Cloudant Query documentation, with references to text indices removed.

PR #122, opened 2017-04-06
Additions to _find and _index from original Mango docs PR

lornajane:
This replaces #31 which added the Mango docs, but mostly it adds anything from that PR that wasn't already covered in #55 so this version now includes more examples and words in places.  I've also tried to fix at least as many indentation weirdnesses as I've created!

One major difference was that #31 covered indexes with `type=text` which these docs didn't - I've added this content in but I wanted to flag this just in case there's a reason why it shouldn't be included in this context.

### DIFF
--- a/src/api/database/find.rst
+++ b/src/api/database/find.rst
@@ -44,18 +44,39 @@
     :query json use_index: Instruct a query to use a specific index. Specified
         either as ``"<design_document>"`` or
         ``["<design_document>", "<index_name>"]``. *Optional*
+    :query number r: Read quorum needed for the result. This defaults to 1, in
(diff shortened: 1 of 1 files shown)


lornajane:
Is there an easy way to run those syntax checks locally, please?  The build works OK but I didn't realise there was more I should have checked.  (we need a contributing file, I've got it on my todo list!)

wohali:
Yes - run `make check`.

lornajane:
Let me know if I should squash these commits, I fixed one check and managed to break another!

wohali:
Yes, please squash! 

----- Original Message -----

wohali:
+1, thanks!

PR #124, opened 2017-04-07
Document filtered replication

flimzy:


wohali:
Can you fix the build please?

https://travis-ci.org/apache/couchdb-documentation/jobs/219876841

You can run this yourself with `make check`.

flimzy:
I'm not sure how to solve this.  I've put the long URL on a line by itself, but 'make check' still complains:

    line 365 : too long (81 > 80) line
          <https://wiki.apache.org/couchdb/Replication#Named_Document_Replication>`_)

If I un-indent that line to get in under the 80-char limit, `make` complains:

    /home/jonhall/git/couchdb-documentation/src/api/server/common.rst:363: WARNING: Inline interpreted text or phrase reference start-string without end-string.

Is there some way I don't know about to use a long URL like that?

wohali:
So three comments, now that I've actually read the PR :sheep: 😀 

1. We're really trying to deprecate the old wiki (wiki.apache.org). We do have a new wiki at https://cwiki.apache.org/confluence/display/COUCHDB/ that everything is supposed to migrate to (and could use your help!) but the overriding principle is that everything *should* be in the docs if it makes sense for it to be there. And this clearly looks like a topic for the documentation, *not* for the wiki. Wiki should be reserved for things like project governance, process and procedure.
2. You can create a short link with https://s.apache.org/ which should last ~forever.
3. It doesn't help for this link but there is another way of doing it, namely:

```
External hyperlinks, like Python_.

.. _Python: http://www.python.org/ 
```

flimzy:
Thanks, I'll work on making sure the relevant info is directly in the documentation.

wohali:
Ping, @flimzy is this ready to go?

flimzy:
@wohali Yes, I think it's ready now.

PR #139, opened 2017-07-04
POST /dbname should not return ETag response header

rnewson:
as it relates to a different resource.

Note: We used to send ETag for this in previous releases, so it's
technically a regression even though we accidentally fix a spec
violation.

Closes apache/couchdb#620

### DIFF
--- a/src/api/database/common.rst
+++ b/src/api/database/common.rst
@@ -292,7 +292,6 @@
       <api/doc/batch-writes>` Possible values: ``ok``. *Optional*
     :>header Content-Type: - :mimetype:`application/json`
                            - :mimetype:`text/plain; charset=utf-8`
-    :>header ETag: Quoted new document's revision
     :>header Location: Document's URI
     :>json string id: Document ID
     :>json boolean ok: Operation status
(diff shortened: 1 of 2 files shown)


wohali:
You should also document this in https://github.com/apache/couchdb-documentation/blob/master/src/whatsnew/2.0.rst as a breaking change.

You probably also want to say " Closes apache/couchdb#620 " since this repo doesn't have an issue #620. :)

wohali:
Oh, +1 once the 2.0.rst is updated.

rnewson:
you mean whatsnew/2.1rst?

rnewson:
2.1.0.rst refers to the scheduler, which isn't present...

wohali:
As discussed on IRC:

```irc
13:49 <+Wohali> rnewson: 2.0 is what you want to put it in, because it's
                changed behaviour in the 2.0 release. We will have to change
                the repo for the scheduler stuff, jan didn't fork that repo
                before it landed. Or we will fork it earlier and cherry pick.
                and make commits ot master that move the scheduler changes to a
                2.2 or similar, unless there is agreement it can land in 2.1 in
                the next couple of weeks
```

wohali:
Taking this one as @rnewson is out of cycles right now.

PR #140, opened 2017-07-05
couchapps -> ddocs

wohali:
* remove all references to couchapps in the documentation
* include an explanation as to why this change has been made
## Checklist

- [ ] Setup redirects at RTD so that hits to http://docs.couchdb.org/en/*/couchapp/* go to a useful place (because couchapp.org redirects to http://docs.couchdb.com/en/latest/couchapp/ddocs.html)


janl:
+100000000000

wohali:
Redirects done.

PR #144, opened 2017-07-14
Update confusing note on dd copy operation

calonso:
## Overview

Just replace a misleading note regarding `COPY` operation on Design documents.


janl:
Just needs trailing whitespace removed before we can merge

janl:
thanks!

PR #148, opened 2017-07-23
Find API uses JSON body, not query parameters

flimzy:
Also update the data type for a few fields, and mark `selector` as required.

## Overview

This corrects the documentation for the _find endpoint to reflect that the parameters must be sent via the JSON body, rather than as query parameters.

## Checklist

### DIFF
--- a/src/api/database/find.rst
+++ b/src/api/database/find.rst
@@ -28,40 +28,40 @@
 
     :<header Content-Type: - :mimetype:`application/json`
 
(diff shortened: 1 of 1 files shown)


wohali:
+1 by inspection: https://github.com/apache/couchdb/blob/master/src/mango/test/mango.py#L156-L185

wohali:
@flimzy You have the power now to merge these with the big green button...but since we're about to release 2.1, I'll press it for you. :D Cheers!

PR #158, opened 2017-08-16
add docs around pagination

garrensmith:


src/api/database/find.rst
```diff
@@ -55,8 +55,7 @@
         returns an opaque string under the ``bookmark`` key that can then be
         passed back in a query to get the next page of results. If any part of
         the query other than ``bookmark`` changes between requests, the results
-        are undefined. *Optional, default: null  Only for indexes of type
-        ``text``.*
+        are undefined. *Optional, default: null
```
willholley: is it just the `selector` that should be consistent between calls to `_find`? If `limit` or `skip` are defined when passing a bookmark, won't that generate unexpected results?

src/api/database/find.rst
```diff
@@ -833,6 +832,18 @@ Example of selective retrieval of fields from matching documents:
...
+==========
+
+Mango queries support pagination via the bookmark field. Each query will
+return a bookmark. A bookmark is a placeholder that CouchDB uses to
+determine where to start the query from. To get the next set of query
+results for a given query, add the bookmark that was received in the previous
+query to your next query. Remember to keep the query selector the same,
+otherwise you will receive unexpected results. This will return the next
+set of results. To paginate backwards use the previous bookmarks to return
+the previous set of results.
```
willholley: probably worth mentioning that the presence of a bookmark doesn't guarantee that there are more results. The only way to test for more results is to fetch the next page.

willholley:
lgtm

garrensmith:
Merged. Thanks @willholley for the help

PR #167, opened 2017-09-27
httpd->chttpd for 2.x proxy auth and SSL

wohali:
We fixed this in https://issues.apache.org/jira/browse/COUCHDB-3203 but never updated the docs to match.

### DIFF
--- a/src/config/http.rst
+++ b/src/config/http.rst
@@ -34,21 +34,6 @@ HTTP Server Options
 
         .. _JSONP: http://www.json-p.org/
 
-    .. config:option:: authentication_handlers :: Authentication handlers
-
-        List of used authentication handlers that used by CouchDB. You may
-        extend them via third-party plugins or remove some of them if you won't
-        let users to use one of provided methods::
-
-            [httpd]
-            authentication_handlers = {couch_httpd_auth, cookie_authentication_handler}, {couch_httpd_auth, default_authentication_handler}
-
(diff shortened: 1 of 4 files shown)


src/whatsnew/1.6.rst
```diff
@@ -29,7 +29,7 @@ The :ref:`Proxy Authentication <api/auth/proxy>` handler was renamed to
 ``proxy_authentication_handler`` to follow the ``*_authentication_handler`` form
 of all other handlers. The old ``proxy_authentification_handler`` name is marked
 as deprecated and will be removed in future releases. It's strongly recommended
-to update :config:option:`httpd/authentication_handlers` option with new value
+to update ``httpd/authentication_handlers`` option with new value
```
wohali: I had to de-reference this because the 1.6 config option no longer exists verbatim, and it would be wrong to link it to the newer option that doesn't exist back in 1.6.

wohali:
@flimzy any chance of reviewing this one too?

src/config/http.rst
```diff
@@ -257,6 +242,21 @@ HTTP Server Options
             the CouchDB server header is replaced with the
             Mochiweb server header.
 
+    .. config:option:: authentication_handlers :: Authentication handlers
+
+        List of used authentication handlers that used by CouchDB. You may
```
flimzy: s/List of used authentication/List of authentication/

wohali:
@flimzy sorry to nag, is that a +1?

PR #175, opened 2017-10-15
Correct document to database

sdaves:
## Overview

Based on view behavior, does a post to /db/_index recalculate all design docs in the database, or just the one design doc?

> View index rebuilds occur when one view from the same the view group (i.e. all the views defined within a single a design document) has been determined as needing a rebuild. For example, if if you have a design document with different views, and you update the database, all three view indexes within the design document will be updated. 

> http://docs.couchdb.org/en/2.1.0/api/ddoc/views.html

## Checklist

### DIFF
--- a/src/api/database/find.rst
+++ b/src/api/database/find.rst
@@ -929,7 +929,7 @@ built using MapReduce Views.
         document.
         Indexes can be grouped into design documents for efficiency. However, a
         change to one index in a design document will invalidate all other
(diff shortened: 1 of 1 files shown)


wohali:
Only the one design doc.

wohali:
https://github.com/apache/couchdb/blob/master/src/mango/src/mango_httpd.erl#L84-L110

PR #178, opened 2017-10-19
Minor changes

pdxfixit:
Just a few minor corrections.

### DIFF
--- a/src/api/database/changes.rst
+++ b/src/api/database/changes.rst
@@ -685,7 +685,7 @@ amount of duplicated code.
     While :ref:`map functions <mapfun>` doesn't process the design documents,
     using ``_view`` filter forces them to do this. You need to be sure, that
     they are ready to handle documents with *alien* structure without panic
-    crush.
+    crash.
 
 .. note::
     Using ``_view`` filter doesn't queries the view index files, so you cannot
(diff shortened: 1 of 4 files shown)


src/api/database/changes.rst
```diff
@@ -685,7 +685,7 @@ amount of duplicated code.
     While :ref:`map functions <mapfun>` doesn't process the design documents,
     using ``_view`` filter forces them to do this. You need to be sure, that
     they are ready to handle documents with *alien* structure without panic
-    crush.
+    crash.
```
flimzy: I'd just remove the word entirely.  "panic crash" is redundant.  I think "without panic." is better. What do you think?

src/config/logging.rst
```diff
@@ -90,7 +90,7 @@ Logging options
         - ``warning`` or ``warn``: Warning messages are alerts about edge situations that
           may lead to errors. For instance, compaction daemon alerts about low
           or insufficient disk space at this level.
-        - ``error`` or ``err``: Error level includes only things that going wrong, crush
+        - ``error`` or ``err``: Error level includes only things that going wrong, crash
```
flimzy: should also replace "going wrong" with "go wrong" while you're there.

src/api/database/changes.rst
pdxfixit: Agreed.

PR #184, opened 2017-10-23
Spelling improvements

flimzy:
This PR represents the result of doing an automated spell-check on the entire `src/` directory, then manually doing corrections.

A number of the "spelling errors" were simply a difference between US and British spelling, so I did a very superficial audit of the documentation (i.e. `grep favorite` and `grep favourite`) and found roughly a 3:1 ratio in preference to American spelling, so chose to standardize on that.  I made that commit separately, though, in case it proves to be at all controversial, to make it easy to remove.

Other than spelling, this contains a couple of other minor improvements and corrections I found while doing the spelling corrections, but tried not to change anything of substance here.

### DIFF
--- a/src/api/basics.rst
+++ b/src/api/basics.rst
@@ -309,7 +309,7 @@ Number Handling
 ---------------
 
 Developers and users new to computer handling of numbers often encounter
-suprises when expecting that a number stored in JSON format does not
+surprises when expecting that a number stored in JSON format does not
 necessarily return as the same number as compared character by character.
 
 Any numbers defined in JSON that contain a decimal point or exponent will be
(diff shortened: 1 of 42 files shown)


janl:
Not a fan of standardising on US spelling. If anything, UK pls. 🇬🇧

rnewson:
same 🇬🇧

rnewson:
the other changes are most welcome though. (and note that there's some leakage between the commits)

flimzy:
> Not a fan of standardising on US spelling. If anything, UK pls. :gb:

I don't have a preference either way. I only chose American spelling simply because it appeared to be the smallest delta.  Should I go the other way and standardize everything to British spelling?

flimzy:
Since :us: vs :gb: spelling turned out to be a matter of concern, I have removed that commit from this PR, to be revisited (probably in reverse) in a later PR.

janl:
nice work

wohali:
@flimzy remember to squash when merging plz and ty :)

flimzy:
> @flimzy remember to squash when merging plz and ty :)

Are you suggesting squashing before merging, or doing a "squash and merge"?  If the latter is the preference for this project, I believe it can be made the only option in the repo settings.

wohali:
The former for a PR that doesn't have distinct commits that should remain distinct. We don't force the latter in the repo settings because, sometimes, rebase and merge makes more sense (where we want to preserve the individual commits in the PR)

PR #186, opened 2017-10-26
Clarify usage of COPY

flimzy:
The documentation suggests that COPY can be used to overwrite the target document, but there's no (obvious) way to actually do this.  The `rev` parameter is used for the source, and there is no (documented/obvious) way to specify the rev for the target document to update. This leads me to believe it is mis-documented.

If someone can explain that my assumption is incorrect, I'll instead update the docs to explain how to overwrite the target.


flimzy:
Ahh thanks! I think it's worth mentioning that in the description of the `Destination` header. I may submit a different PR for that later.

PR #193, opened 2017-11-01
Updates for 2.1.1-RC2

wohali:


src/whatsnew/2.1.rst
```diff
@@ -46,6 +46,15 @@ Upgrade Notes
   had been unfortunately misnamed, and has now been updated to behave as the
   name would suggest. Both are documented in the shipped ``default.ini`` file.
 
+* :ghissue:`914`: Certain critical config sections are blacklisted from being
```
flimzy: Might be nice to mention these sections here, if they're not too esoteric.

davisp:
Belated +1

src/whatsnew/2.1.rst
wohali: They're pretty esoteric - no one should be accessing these over the HTTP API.

PR #194, opened 2017-11-01
Update example replication document to reflect 2.1

flimzy:
We're still using old (2.0?) replication document example in one place I found. This updates it.

The other documentation I found relating to these documents appears to be up to date.

### DIFF
--- a/src/replication/replicator.rst
+++ b/src/replication/replicator.rst
@@ -73,18 +73,24 @@ Replication state of this document can then be queried from
 .. code-block:: json
 
      {
-         "database": "_replicator",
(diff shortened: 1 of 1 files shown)


wohali:
FYI there are a lot of sample documents that will be outdated in the documentation.

Perhaps we can consider some sort of dynamic generation of these examples with each release...

flimzy:
> Perhaps we can consider some sort of dynamic generation of these examples with each release...

I like that idea. I'm not sure how best to do it; that could make for a pretty hefty build-dependency.

PR #199, opened 2017-11-07
Clarify change to default node name in 2.1.1

wohali:
user@ reported that the change to the default node name tripped them up. Would be wise to add the remediation steps to the release notes.


janl:
+1

PR #219, opened 2017-12-12
Fix _cluster_setup api | couchdb node name changes | verify cluster setup  

raghavtan:
- When doing cluster_setup via api or wizard we need to make changes in vm.args as all nodes should have a unique name (mentioned in the file comments but not in the cluster setup documentation)
- executing _cluster_setup api with enable_cluster the second time gives error response on the coordination node when adding nodes to cluster hence is not required causes failure
- only add_node action is required from _cluster_setup api on  coordination node
- Added cluster setup verification docs

### DIFF
--- a/src/cluster/setup.rst
+++ b/src/cluster/setup.rst
@@ -167,6 +167,11 @@ Before you can add nodes to form a cluster, you have to have them
 listen on a public IP address and set up an admin user. Do this, once
 per node:
 
+.. note::
+change line ``-name couchdb@127.0.0.1`` in file /Couch-Install-Home/etc/vm.args
+on each node for clustered setup, each node in the system must have a unique name
+eg. ``-name couchdb@couch1`` and ``-name couchdb@couch2``
+
 .. code-block:: bash
 
     curl -X PUT http://127.0.0.1:5984/_node/couchdb@<this-nodes-ip-address>/_config/admins/admin -d '"password"'
@@ -211,12 +216,11 @@ requires all other nodes to be able to see it and vice versa.
 Set up will not work with unavailable nodes.
 The notion of "setup coordination node" will be gone once the setup is finished.
 From then on, the cluster will no longer have a "setup coordination node".
-To add a node run these two commands:
+To add a node run this commands for each node you want to add:
 
 .. code-block:: bash
 
-    curl -X POST -H "Content-Type: application/json" http://admin:password@127.0.0.1:5984/_cluster_setup -d '{"action": "enable_cluster", "bind_address":"0.0.0.0", "username": "admin", "password":"password", "port": 15984, "node_count": "3", "remote_node": "<remote-node-ip>", "remote_current_user": "<remote-node-username>", "remote_current_password": "<remote-node-password>" }'
-    curl -X POST -H "Content-Type: application/json" http://admin:password@127.0.0.1:5984/_cluster_setup -d '{"action": "add_node", "host":"<remote-node-ip>", "port": "<remote-node-port>", "username": "admin", "password":"password"}'
+    curl -X POST -H "Content-Type: application/json" http://admin:password@127.0.0.1:5984/_cluster_setup -d '{"action": "add_node", "host":"<remote-node-ip>", "port": <remote-node-port>, "username": "admin", "password":"password"}'
 
 This will join the two nodes together.
 Keep running the above commands for each
@@ -226,6 +230,29 @@ following command to complete the setup and add the missing databases:
 .. code-block:: bash
 
     curl -X POST -H "Content-Type: application/json" http://admin:password@127.0.0.1:5984/_cluster_setup -d '{"action": "finish_cluster"}'
+
+Verify install
+.. code-block:: bash
+    curl http://admin:password@127.0.0.1:5984/_cluster_setup
+Response
+.. code-block:: bash
+    {"state":"cluster_finished"}
+
+Verify cluster nodes
+.. code-block:: bash
+    curl http://admin:password@127.0.0.1:5984/_membership
+Response
+.. code-block:: bash
+    {
+    "all_nodes": [
+        "couchdb@couch1",
+        "couchdb@couch2",
+    ],
+    "cluster_nodes": [
+        "couchdb@couch1",
+        "couchdb@couch2",
+    ]
+}
 
 You CouchDB cluster is now set up.
 
--- a/src/cluster/setup.rst
+++ b/src/cluster/setup.rst
@@ -167,9 +167,9 @@ Before you can add nodes to form a cluster, you have to have them
 listen on a public IP address and set up an admin user. Do this, once
 per node:
 
-Change line ``-name couchdb@127.0.0.1`` in file /Couch-Install-Home/etc/vm.args
-on each node for clustered setup, each node in the system must have a unique name.
-eg. ``-name couchdb@couch1`` and ``-name couchdb@couch2``
+Change line ``-name couchdb@127.0.0.1`` in file /Couch-Home/etc/vm.args on
+each node for clustered setup, each node in system must have a unique name.
+eg. ``-name couchdb@couch1`` | ``-name couchdb@couch2``
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


src/cluster/setup.rst
```diff
@@ -167,6 +167,10 @@ Before you can add nodes to form a cluster, you have to have them
 listen on a public IP address and set up an admin user. Do this, once
 per node:
 
+Change line ``-name couchdb@127.0.0.1`` in file /Couch-Home/etc/vm.args on
```
wohali: Just specify the path here as `etc/vm.args`, please.

src/cluster/setup.rst
```diff
@@ -167,6 +167,10 @@ Before you can add nodes to form a cluster, you have to have them
...
+each node for clustered setup, each node in system must have a unique name.
+eg. ``-name couchdb@couch1`` | ``-name couchdb@couch2``
```
wohali: The `couch1`, `couch2` hostnames must be fully qualified domain names (FQDNs) with valid DNS entries, see [the comment in vm.args](https://github.com/apache/couchdb/blob/master/rel/overlay/etc/vm.args#L13-L28) for more detail.

src/cluster/setup.rst
```diff
@@ -211,12 +215,11 @@ requires all other nodes to be able to see it and vice versa.
 Set up will not work with unavailable nodes.
 The notion of "setup coordination node" will be gone once the setup is finished.
 From then on, the cluster will no longer have a "setup coordination node".
-To add a node run these two commands:
+To add a node run this commands for each node you want to add:
 
 .. code-block:: bash
 
-    curl -X POST -H "Content-Type: application/json" http://admin:password@127.0.0.1:5984/_cluster_setup -d '{"action": "enable_cluster", "bind_address":"0.0.0.0", "username": "admin", "password":"password", "port": 15984, "node_count": "3", "remote_node": "<remote-node-ip>", "remote_current_user": "<remote-node-username>", "remote_current_password": "<remote-node-password>" }'
```
wohali: No, actually, you need this first statement too. This sets the bind address, bind port, admin username and admin password on the remote node. The node_count parameter is also critical.

You can see a test cluster being set up in the test case for the `couch_setup` application [here](https://github.com/apache/couchdb-setup/blob/master/test/t-frontend-setup.sh) and [here](https://github.com/apache/couchdb-setup/blob/master/test/t.sh).

src/cluster/setup.rst
```diff
@@ -211,12 +215,11 @@ requires all other nodes to be able to see it and vice versa.
...
-    curl -X POST -H "Content-Type: application/json" http://admin:password@127.0.0.1:5984/_cluster_setup -d '{"action": "add_node", "host":"<remote-node-ip>", "port": "<remote-node-port>", "username": "admin", "password":"password"}'
+    curl -X POST -H "Content-Type: application/json" http://admin:password@127.0.0.1:5984/_cluster_setup -d '{"action": "add_node", "host":"<remote-node-ip>", "port": <remote-node-port>, "username": "admin", "password":"password"}'
```
wohali: You're also missing a quote mark here. The original line was correct.

src/cluster/setup.rst
```diff
@@ -227,6 +230,39 @@ following command to complete the setup and add the missing databases:
...
+        "all_nodes": [
+            "couchdb@couch1",
+            "couchdb@couch2",
+        ],
+        "cluster_nodes": [
+            "couchdb@couch1",
+            "couchdb@couch2",
+        ]
+    }
+
```
wohali: This is a useful thing to add, thanks.

src/cluster/setup.rst
```diff
@@ -167,6 +167,9 @@ Before you can add nodes to form a cluster, you have to have them
 listen on a public IP address and set up an admin user. Do this, once
 per node:
 
+Change line ``-name couchdb@127.0.0.1`` in file etc/vm.args on
+each node for clustered setup, each node in system must have a unique name.
```
flimzy: This is a run-on sentence. I suggest changing to:

> ... for clustered setup. Each node in the system must have a unique name.

src/cluster/setup.rst
```diff
@@ -211,12 +214,11 @@ requires all other nodes to be able to see it and vice versa.
...
 The notion of "setup coordination node" will be gone once the setup is finished.
 From then on, the cluster will no longer have a "setup coordination node".
-To add a node run these two commands:
+To add a node run these commands for each node you want to add:
 
 .. code-block:: bash
-
     curl -X POST -H "Content-Type: application/json" http://admin:password@127.0.0.1:5984/_cluster_setup -d '{"action": "enable_cluster", "bind_address":"0.0.0.0", "username": "admin", "password":"password", "port": 15984, "node_count": "3", "remote_node": "<remote-node-ip>", "remote_current_user": "<remote-node-username>", "remote_current_password": "<remote-node-password>" }'
-    curl -X POST -H "Content-Type: application/json" http://admin:password@127.0.0.1:5984/_cluster_setup -d '{"action": "add_node", "host":"<remote-node-ip>", "port": "<remote-node-port>", "username": "admin", "password":"password"}'
+    curl -X POST -H "Content-Type: application/json" http://admin:password@127.0.0.1:5984/_cluster_setup -d '{"action": "add_node", "host":"<remote-node-ip>", "port": <remote-node-port>, "username": "admin", "password":"password"}'
```
flimzy: Are you sure the port doesn't need to be a string? AFAIK, all config values are strings. Not to mention, a port can (probably) be non-numeric, i.e. `http` or `https`, etc.

raghavtan: @flimzy sure of that the port doesnt need to be string.

src/cluster/setup.rst
raghavtan: np

src/cluster/setup.rst
```diff
@@ -167,6 +167,9 @@ Before you can add nodes to form a cluster, you have to have them
 listen on a public IP address and set up an admin user. Do this, once
 per node:
 
+``-name couchdb@127.0.0.1`` in file etc/vm.args on
```
flimzy: Now this is an incomplete sentence :)

src/cluster/setup.rst
```diff
@@ -164,9 +164,15 @@ procedure on other machines. Be sure to specify the total number of nodes you
 expect to add to the cluster before adding nodes.
 
 Before you can add nodes to form a cluster, you have to have them
-listen on a public IP address and set up an admin user. Do this, once
+listen on a node address if all nodes can communicate over those IP's,
+also set up an admin user which should be same on all nodes. Do this, once
 per node:
```
flimzy: This is some very awkward wording, IMO. I suggest rewriting the paragraph as:

> Before you can add nodes to form a cluster, you must have them listening on an IP address accessible from the other nodes in the cluster.  Do this once per node:

> In file ``etc/vm.args`` change the the line ``-name couchdb@127.0.0.1`` to ``-name couchdb@<this-nodes-ip-address>`` for each node. For clustered setup, each node in the system must have a unique name.

raghavtan: Rephrased but added parameter <FQDN> along with IP @flimzy

src/cluster/setup.rst
```diff
@@ -163,9 +163,17 @@ then to add nodes by IP address. To get more nodes, go through the same install
...
 
-Before you can add nodes to form a cluster, you have to have them
-listen on a public IP address and set up an admin user. Do this, once
-per node:
+Before you can add nodes to form a cluster, you must have them listening on an
+IP address accessible from the other nodes in the cluster.
+Do this once per node:
+
+In file etc/vm.args  to -name couchdb@<this-nodes-ip-address> for each node.
+For clustered setup, each node in the system must have a unique name.
```
flimzy: This paragraph looks like a partial copy of the following one.  Copy and paste error?

wohali: @raghavtan can you check this comment from @flimzy before I squash and merge? thanks.

raghavtan:
@wohali fixed

wohali:
@raghavtan Thanks again, and congratulations on your first contribution to Apache CouchDB!

PR #227, opened 2018-01-06
Document changes to the `couchdb` start script

efine:
## Overview

The existing `couchdb` start script hard-codes the arguments to
`-args_file` and `-config`. Although it is possible to copy this
script and modify it, or modify it in place, that is less than ideal
and can lead to all kinds of difficulties.

This PR documents the user-facing capabilities introduced by the corresponding PR on the `couchdb` repository.

## Related Issues or Pull Requests

apache/couchdb/pull/1095


kocolosk:
Definitely useful, I had wanted to do this a few months back but never got around to it. Thanks!

PR #238, opened 2018-01-23
added header info to replication api

drunksaint:
## Overview

Replication between 2 couchdb nodes with custom headers is implemented, but not present in the documentation. This commit adds that.

## Checklist


janl:
Thanks!

PR #247, opened 2018-02-14
Grammar corrections

flimzy:


src/cluster/nodes.rst
```diff
@@ -22,7 +22,7 @@ Adding a node
 =============
 
 Go to ``http://server1:5984/_membership`` to see the name of the node and all
-the nodes it knows about and are connected too.
+the nodes it knows about and is connected to.
```
wohali: Why not also remove the dangling participle? :)

wohali:
+1

src/cluster/nodes.rst
flimzy: Touché!

I suppose

> ... to see the name of the node and the nodes it is connected to and knows about.

would be an improvement, eh?

PR #252, opened 2018-02-27
Suggested edit to make clear what a quorum actually represents.

oliverdain:
## Overview

I found the documentation about read and write quorums for a cluster mode DB confusing. I think this edit makes things a bit more clear (though I don't love the wording).

This came from a Slack conversation which you can find here if curious: https://couchdb.slack.com/archives/C49LEE7NW/p1519761842000347

### DIFF
--- a/src/cluster/theory.rst
+++ b/src/cluster/theory.rst
@@ -34,10 +34,17 @@ thereby override the defaults in ``default.ini``.
 
 The number of copies of a document with the same revision that have to be read
 before CouchDB returns with a ``200`` is equal to a half of total copies of
-the document plus one. It is the same for the number of nodes that need to
-save a document before a write is returned with ``201``. If there are less
-nodes than that number, then ``202`` is returned. Both read and write numbers
-can be specified with a request as ``r`` and ``w`` parameters accordingly.
+the document plus one *in the current partition*. That is, if there is a network
+partition and, for example, only 1 out of 3 nodes is available you can still
+get a ``200`` response but the data will not have come from a true quorum and
+might be stale.
+
+The number of nodes that need to save a document before a write is returned with
+``201`` is also equal to half of the total copies of the document plus one.
+However, unlike the read case, for writes this is computed for the database as a
+whole, even if there is a network partition. If there are less nodes than that
+number, then ``202`` is returned. Both read and write numbers can be specified
+with a request as ``r`` and ``w`` parameters accordingly.
 
 We will focus on the shards and replicas for now.
 
--- a/src/cluster/theory.rst
+++ b/src/cluster/theory.rst
@@ -32,10 +32,10 @@ As you see in ``etc/default.ini`` there is a section called [cluster]
 When creating a database you can send your own values with request and
 thereby override the defaults in ``default.ini``.
 
-In clustered operation, a quorum must be reached before CouchDB returns a 200
-for a fetch, or 201 for a write operation. A quorum is defined one plus half the
-number of "relevant copies". "Relevant copies" is defined slightly differently
-for read and write operations.
+In clustered operation, a quorum must be reached before CouchDB returns a
+``200`` for a fetch, or 201 for a write operation. A quorum is defined as one
+plus half the number of "relevant copies". "Relevant copies" is defined
+slightly differently for read and write operations.
 
 For read operations, the number of relevant copies is the number of
 currently-accessible shards holding the requested data, meaning that in the case
@@ -44,7 +44,7 @@ than the number of replicas in the cluster.  The number of read copies can be
 set with the rparameter.
 
 For write operations the number of relevant copies is always `n`, the number of
-replicas in teh cluster.  For write operations, the number of copies can be set
+replicas in the cluster.  For write operations, the number of copies can be set
 using the w parameter. If fewer than this number of nodes is available, a 202
 will be returned.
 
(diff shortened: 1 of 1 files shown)


src/cluster/theory.rst
```diff
@@ -34,10 +34,17 @@ thereby override the defaults in ``default.ini``.
 
 The number of copies of a document with the same revision that have to be read
 before CouchDB returns with a ``200`` is equal to a half of total copies of
-the document plus one. It is the same for the number of nodes that need to
-save a document before a write is returned with ``201``. If there are less
-nodes than that number, then ``202`` is returned. Both read and write numbers
-can be specified with a request as ``r`` and ``w`` parameters accordingly.
+the document plus one *in the current partition*. That is, if there is a network
```
flimzy: "partition" is not a word common to CouchDB parlance. Do you mean node? shard?

src/cluster/theory.rst
```diff
@@ -34,10 +34,17 @@ thereby override the defaults in ``default.ini``.
...
+partition and, for example, only 1 out of 3 nodes is available you can still
+get a ``200`` response but the data will not have come from a true quorum and
+might be stale.
+
+The number of nodes that need to save a document before a write is returned with
+``201`` is also equal to half of the total copies of the document plus one.
+However, unlike the read case, for writes this is computed for the database as a
+whole, even if there is a network partition. If there are less nodes than that
```
flimzy: "less" -> "fewer" to satisfy the grammar nazis

src/cluster/theory.rst
oliverdain: I was trying to describe a network partition. Specifically, here I'm trying to say that if there's a network partition then the quorum is relative to only the subset of nodes on the side of the partition where the client is.

Definitely open to different wording if this is confusing.

src/cluster/theory.rst
oliverdain: Will fix.

oliverdain:
Thanks for the review @flimzy . Made the obvious grammar fix. Not entirely sure how to fix the 1st paragraph. Would love suggestions.

src/cluster/theory.rst
flimzy: Ah, now I see what you're getting at.  That is an important point. And the paragraph was already a bit confusing before your addition. Perhaps a rewrite of the entire section is in order.

Example:

> In clustered operation, a quorum must be reached before CouchDB returns a ``200`` for a fetch, or ``201`` for a write operation. A quorum is defined one plus half the number of copies.

> For read operations, the maximum number of copies is the number of currently-accessible nodes, meaning that in the case of a failure or network partition, the effective number of copies may be lower.  The number of read copies can be set with the ``r``parameter.

> For write operations, the number of copies can be set using the ``w`` parameter. If fewer than this number of nodes is available, a ``202`` will be returned.

oliverdain: Overall I like what you have better. Will start with that but make a few changes to be more explicit. Updated PR coming soon...

oliverdain:
OK. Ready for re-review. Thanks @flimzy

src/cluster/theory.rst
```diff
@@ -32,12 +32,21 @@ As you see in ``etc/default.ini`` there is a section called [cluster]
...
 thereby override the defaults in ``default.ini``.
 
-The number of copies of a document with the same revision that have to be read
-before CouchDB returns with a ``200`` is equal to a half of total copies of
-the document plus one. It is the same for the number of nodes that need to
-save a document before a write is returned with ``201``. If there are less
-nodes than that number, then ``202`` is returned. Both read and write numbers
-can be specified with a request as ``r`` and ``w`` parameters accordingly.
+In clustered operation, a quorum must be reached before CouchDB returns a 200
+for a fetch, or 201 for a write operation. A quorum is defined one plus half the
```
flimzy: `A quorum is defined as one...`

src/cluster/theory.rst
```diff
@@ -32,12 +32,21 @@ As you see in ``etc/default.ini`` there is a section called [cluster]
 When creating a database you can send your own values with request and
 thereby override the defaults in ``default.ini``.
 
-The number of copies of a document with the same revision that have to be read
-before CouchDB returns with a ``200`` is equal to a half of total copies of
-the document plus one. It is the same for the number of nodes that need to
-save a document before a write is returned with ``201``. If there are less
-nodes than that number, then ``202`` is returned. Both read and write numbers
-can be specified with a request as ``r`` and ``w`` parameters accordingly.
+In clustered operation, a quorum must be reached before CouchDB returns a 200
```
flimzy: Probably good idea to wrap status codes with tick marks as before:

    ``200``

vs

    200

src/cluster/theory.rst
```diff
@@ -32,12 +32,21 @@ As you see in ``etc/default.ini`` there is a section called [cluster]
...
+for read and write operations.
+
+For read operations, the number of relevant copies is the number of
+currently-accessible shards holding the requested data, meaning that in the case
+of a failure or network partition, the number of relevant copies may be lower
+than the number of replicas in the cluster.  The number of read copies can be
+set with the rparameter.
+
+For write operations the number of relevant copies is always `n`, the number of
+replicas in teh cluster.  For write operations, the number of copies can be set
```
flimzy: teh -> the

oliverdain:
All fixed. Thanks for the review.

flimzy:
+1

oliverdain:
@flimzy what's your process? Do you do the merge or should I do that?

Thanks again for the review.

oliverdain:
Hey @flimzy thanks again for your review here. I noticed the live docs haven't changed yet (http://docs.couchdb.org/en/2.1.1/cluster/theory.html). I kinda thought when this was merged the docs would auto-update. Is that not the case?

PR #253, opened 2018-03-05
Remove mention of Fauxton on port 5986

wohali:
See apache/couchdb#1198 for detail.

Even without apache/couchdb#1199 landing, this will at least stop directing people to use a feature we never intended to make available.

This whole page needs a rewrite; @garbados has agreed to take a look in the future. In case that doesn't land for 2.2.0, I wanted to get this quick fix in place instead.


davisp:
+1

PR #256, opened 2018-03-17
Fix whitespace in bulleted lists

### LINKED ISSUE #201: [theme] fix whitespace in bulleted lists
http://docs.couchdb.org/en/2.1.1/whatsnew/2.1.html This is kind of unreadable. We need newline breaks between each bullet point. Is it possible to do it while still keeping our dependency on the upstream?

wohali:
Fixes #201.

Note that, in my testing, prefixing the list with:

```rst
.. rst-class:: open
```

may be necessary to force the double-spacing, especially if the list is not internally double-spaced already, or has lines that are only one line long.

### DIFF
--- a/static/css/rtd_theme.css
+++ b/static/css/rtd_theme.css
@@ -47,3 +47,10 @@ span.linkdescr {
     font-size: 90%;
 }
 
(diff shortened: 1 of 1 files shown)


wohali:
As this is a cosmetic fix, I'm taking the liberty to merge this without a +1.

PR #258, opened 2018-03-17
Document /_up endpoint

### LINKED ISSUE #207: Document the /_up endpoint
Should be in the HTTP API reference. It's not.

wohali:
Fixes #207.

### DIFF
--- a/src/api/server/common.rst
+++ b/src/api/server/common.rst
@@ -1407,6 +1407,40 @@ is as follows:
     :>header Last-Modified: Static files modification timestamp
     :code 200: Request completed successfully
 
+.. _api/server/up:
+
+========
+``/_up``
+========
+
+.. versionadded:: 2.0
+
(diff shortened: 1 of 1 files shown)


src/api/server/common.rst
```diff
@@ -1452,6 +1452,41 @@ is as follows:
...
+
+.. versionadded:: 2.0
+
+.. http:get:: /_up
+    :synopsis: Health check endpoint
+
+    Confirms that the server is up, running, and ready to respond to requests.
+    If the server is in ``true`` or ``nolb``
+    :config:option:`maintenance_mode <couchdb/maintenance_mode>`, the endpoint
+    will return a 404 response.
```
flimzy: I wonder if something like this would be more clear?

> If maintenance mode is ``true`` or ``nolb``, the endpoint will return a 404 response.

PR #263, opened 2018-03-19
Add the POST _bulk_get endpoint

### LINKED ISSUE #246: Add documentation about the POST _bulk_get endpoint
Hello, the POST _bulk_get endpoint is implemented by CouchDB since version 2.0.0 (at least, this is what the release notes say), but I don't find any documentation about it on [the API reference](http://docs.couchdb.org/en/2.1.1/api/index.html). It'd be nice to document it.

I think it follows the same format as described in https://github.com/couchbase/sync_gateway/wiki/Bulk-GET.

nono:
## Overview

The POST _bulk_get endpoint was not documented, and not even listed in the API reference. I have added a short description with an example.

## Testing recommendations

The POST _bulk_get endpoint should appear in the API reference.

## GitHub issue number

Fixes #246

## Checklist

### DIFF
--- a/src/api/database/bulk-api.rst
+++ b/src/api/database/bulk-api.rst
@@ -497,14 +497,14 @@ Sending multiple queries to a database
             "docs": [
                 {
                     "id": "foo"
-                    "_rev": "4-753875d51501a6b1883a9d62b4d33f91",
+                    "rev": "4-753875d51501a6b1883a9d62b4d33f91",
                 },
                 {
                     "id": "foo"
-                    "_rev": "1-4a7e4ae49c4366eaed8edeaea8f784ad",
+                    "rev": "1-4a7e4ae49c4366eaed8edeaea8f784ad",
                 },
                 {
-                    "_id": "bar",
+                    "id": "bar",
                 }
             ]
         }
--- a/src/api/database/bulk-api.rst
+++ b/src/api/database/bulk-api.rst
@@ -465,20 +465,9 @@ Sending multiple queries to a database
 .. http:post:: /{db}/_bulk_get
     :synopsis: Fetches several documents at the given revision
 
-    This method is used by PouchDB and other client to improve the pull
-    replications, by allowing the client to request multiple documents in one
-    request.
-
-    The usual CouchDB API for bulk gets -- a POST to _all_docs with an array of
-    docIDs in the body -- isn't suitable for the replicator because:
-
-    - there's no way to specify which specific revisions are needed (it always
-      returns the default revision),
-    - the response doesn't contain the revision history,
-    - there's no way to specify that only recently changed attachments should
-      be included (a la ``atts_since``),
-    - and attachment bodies can only be encoded inline (base64) not as MIME
-      multipart bodies.
+    This method can be called to query several documents in a bulk. It is well
+    fitted when you want a specific revision of the documents, like replicators
+    do for example, or get the revision history.
 
     :param db: Database name
     :<header Accept: - :mimetype:`application/json`
@@ -490,6 +479,10 @@ Sending multiple queries to a database
     :>json object results: the documents, with the additionnal ``_revisions``
       property that lists the parent revisions if ``revs=true``
     :code 200: Request completed successfully
+    :code 400: The request provided invalid JSON data or invalid query parameter
+    :code 401: Read permission required
+    :code 404: Invalid database name
+    :code 415: Bad :header:`Content-Type` value
 
     **Request**:
 
--- a/src/api/database/bulk-api.rst
+++ b/src/api/database/bulk-api.rst
@@ -456,6 +456,139 @@ Sending multiple queries to a database
     The multiple queries are also supported in /db/_local_docs/queries and
     /db/_design_docs/queries (similar to /db/_all_docs/queries).
 
+.. _api/db/bulk_get:
+
+==================
+``/db/_bulk_get``
+==================
+
+.. http:post:: /{db}/_bulk_get
+    :synopsis: Fetches several documents at the given revision
+
+    This method is used by PouchDB and other client to improve the pull
+    replications, by allowing the client to request multiple documents in one
+    request.
+
+    The usual CouchDB API for bulk gets -- a POST to _all_docs with an array of
+    docIDs in the body -- isn't suitable for the replicator because:
+
+    - there's no way to specify which specific revisions are needed (it always
+      returns the default revision),
+    - the response doesn't contain the revision history,
+    - there's no way to specify that only recently changed attachments should
+      be included (a la ``atts_since``),
+    - and attachment bodies can only be encoded inline (base64) not as MIME
+      multipart bodies.
+
+    :param db: Database name
+    :<header Accept: - :mimetype:`application/json`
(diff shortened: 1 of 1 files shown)


src/api/database/bulk-api.rst
```diff
@@ -456,6 +456,139 @@ Sending multiple queries to a database
...
+==================
+``/db/_bulk_get``
+==================
+
+.. http:post:: /{db}/_bulk_get
+    :synopsis: Fetches several documents at the given revision
+
+    This method is used by PouchDB and other client to improve the pull
+    replications, by allowing the client to request multiple documents in one
+    request.
```
wohali: This is nice info, but the context is wrong. The tone is somewhat out of place as compared to the other endpoints; can you rewrite this to discuss the endpoint without passing value judgments on other endpoints?

src/api/database/bulk-api.rst
```diff
@@ -456,6 +456,139 @@ Sending multiple queries to a database
...
+    The usual CouchDB API for bulk gets -- a POST to _all_docs with an array of
+    docIDs in the body -- isn't suitable for the replicator because:
+
+    - there's no way to specify which specific revisions are needed (it always
+      returns the default revision),
+    - the response doesn't contain the revision history,
+    - there's no way to specify that only recently changed attachments should
+      be included (a la ``atts_since``),
+    - and attachment bodies can only be encoded inline (base64) not as MIME
+      multipart bodies.
```
wohali: Same comment for this section as above. You're kicking sand on `_all_docs` but not describing the specific things that this endpoint does.

Try and use a neutral point of view and describe just what `_bulk_get` does first. Then, compare it with `_all_docs` and explain its unique advantages, if you like.

src/api/database/bulk-api.rst
```diff
@@ -456,6 +456,139 @@ Sending multiple queries to a database
...
+    :param db: Database name
+    :<header Accept: - :mimetype:`application/json`
+    :<header Content-Type: :mimetype:`application/json`
+    :query boolean revs: Give the revisions history
+    :<json array docs: List of document objects, with ``id``, and optionnaly
+      ``rev`` and ``atts_since``
+    :>header Content-Type: - :mimetype:`application/json`
+    :>json object results: the documents, with the additionnal ``_revisions``
+      property that lists the parent revisions if ``revs=true``
+    :code 200: Request completed successfully
```
wohali: Are there other codes it can return? If unsure, check the source code. Here's one possible error it can return: https://github.com/apache/couchdb/blob/master/src/chttpd/src/chttpd_db.erl#L475

src/api/database/bulk-api.rst
```diff
@@ -456,6 +456,139 @@ Sending multiple queries to a database
...
+                        "309651b95df56d52658650fb64257b97"
+                      ]
+                    }
+                  }
+                }
+              ]
+            }
+          ]
+        }
+
```
wohali: Good example.

src/api/database/bulk-api.rst
nono: I'm not sure: is it about this sentence or about the text just below?

src/api/database/bulk-api.rst
nono: OK, I've copied https://github.com/couchbase/sync_gateway/wiki/Bulk-GET#why-is-this-necessary because I didn't know what to write here. I'm going to rewrite it with your advices.

src/api/database/bulk-api.rst
nono: I was not sure what codes should be put here. And, as this endpoint has the same codes than `POST _all_docs`, I've copied the codes from it. I suppose I can make another pull-request to also improve documentation about `POST _all_docs`.

src/api/database/bulk-api.rst
nono: Thanks for the quick review!

src/api/database/bulk-api.rst
nono: To be honest, I don't understand why it was a new endpoint and not just some options for `POST _all_docs`. It looks like they both do the same, except one supports some options and the other some other options. I don't understand the internal of CouchDB, it probably make senses. But I have some difficulties to describe this endpoint.

src/api/database/bulk-api.rst
```diff
@@ -456,6 +456,132 @@ Sending multiple queries to a database
...
+.. _api/db/bulk_get:
+
+==================
+``/db/_bulk_get``
+==================
+
+.. http:post:: /{db}/_bulk_get
+    :synopsis: Fetches several documents at the given revision
+
+    This method can be called to query several documents in a bulk. It is well
```
flimzy: remove the word "a":

> This method can be called to query several documents in bulk.

src/api/database/bulk-api.rst
```diff
@@ -456,6 +456,132 @@ Sending multiple queries to a database
...
+    fitted when you want a specific revision of the documents, like replicators
+    do for example, or get the revision history.
```
flimzy: I suggest a rewording to sound more natural, and to fix some grammatical mistakes:

> It is well suited for fetching a specific revision of documents, as replicators do for example, or for getting revision history.

src/api/database/bulk-api.rst
```diff
@@ -456,6 +456,132 @@ Sending multiple queries to a database
...
     /db/_design_docs/queries (similar to /db/_all_docs/queries).
 
+.. _api/db/bulk_get:
+
+==================
+``/db/_bulk_get``
+==================
+
+.. http:post:: /{db}/_bulk_get
+    :synopsis: Fetches several documents at the given revision
```
flimzy: What do you mean by "at the given revision"?  If I'm not mistaken, this method (like most others) defaults to the current revision, and specifying the revision is optional?  If my understanding is accurate, I suggest simplifying the description to just:

> Fetches several documents simultaneously

nono: It's true, but if you just want to fetch several documents simultaneously, you should probably use `/db/_all_docs` that is simpler.

flimzy: Mmm, that's a fair point. :)

In that case, I suggest simply changing "revision" to "revisions".

nono: It's done

src/api/database/bulk-api.rst
wohali: I would at least add the `400` `bad_request` error that I linked to.

nono: I'm not sure I understand. I've added the line `:code 400: The request provided invalid JSON data or invalid query parameter`, and the invalid JSON data is for this error. Should I rewrite it differently?

wohali: Oh, did you? The code excerpt in the short view in GH doesn't show that. Thanks! This is resolved.

nono: Yes, GH just shows the lines above the comment, not those below. And when there are several code excerpts in a row, it's very easy to be confused (at least, it happens to me sometimes).

wohali:
@flimzy are you satisfied? If so please change your review status.

PR #264, opened 2018-03-22
Improve instructions on upgrading to latest CouchDB releases

wohali:
Some lovely improvements to our upgrade documentation, if I do say so myself.

## Checklist

### DIFF
--- a/src/install/upgrading.rst
+++ b/src/install/upgrading.rst
@@ -12,7 +12,66 @@
 
 .. _install/upgrading:
 
+=====================================
+Upgrading from prior CouchDB releases
+=====================================
+
+Important Notes
+===============
+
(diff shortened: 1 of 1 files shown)


flimzy:
I guess just one comment: You mention that cluster upgrades are automatable, but don't make further mention. I suppose you mean they can be scripted, not that there's some "upgrade cluster" button/featuer that can be activated.  Maybe simply changing "automatable" to "scriptable" would clear up any confusion. It's also possible that there isn't any real confusion to clear up.... just a thought that crossed my mind.

wohali:
@nickva I'll just set them back to "Relax" since that appears everywhere else in the documentation. Thx for the suggestion.

PR #268, opened 2018-04-06
Rewrite sharding documentation

garbados:
## Overview

This WIP PR attempts to revisit the documentation on sharding to make it clearer, and to include detailed instructions for adding and removing nodes and shards.

## Checklist


src/cluster/sharding.rst
```diff
@@ -97,27 +224,49 @@ called ``small`` there. Let us look in it. Yes, you can get it with curl too:
...
+cluster!**
 
-Nothing here, nothing there, a shard in my sleeve
--------------------------------------------------
+To add a shard to a node, add entries like this to the database
+metadata's ``changelog`` attribute:
 
-Start node2 and add it to the cluster. Check in ``/_membership`` that the
-nodes are talking with each other.
+.. code:: json1
```
wohali: This is causing the build to fail, I think the `1` is the problem

wohali:
You say this is a WIP PR - just remove the tag when you're ready for this to be merged.

src/cluster/sharding.rst
```diff
@@ -18,43 +18,170 @@ Sharding
...
+
+Here is a similar example for writing a document:
+
+.. code:: bash
+
+    $ curl -X PUT "$COUCH_URL:5984/{docId}?w=2" -d '{}'
+
+Setting ``r`` or ``w`` to be equal to ``n`` (the number of replicas)
+means you will only receive a response once all nodes with relevant
+shards have responded, however even this does not guarantee `ACIDic
```
wohali: ...or if that number of replicas have been consulted, and timed out while waiting for a response.

This is a good one to try yourself: set up a 3 node cluster, then kill a node, then try to issue a `r=3` or `w=3` request and see what response you get (and how long it take).

src/cluster/sharding.rst
```diff
@@ -18,43 +18,170 @@ Sharding
...
+-------------
+
+To add a node to a cluster, first you must have the additional node
+running somewhere. Make note of the address it binds to, like
+``127.0.0.1``, then ``PUT`` an empty document to the ``/_node``
+endpoint:
+
+.. code:: bash
+
+    $ curl -X PUT "$COUCH_URL:5984/_node/{name}@{address}" -d '{}'
```
wohali: I think this is wrong, don't you mean `127.0.0.1:5986/_nodes/{name}@{address}`, issued on one machine in the cluster?

src/cluster/sharding.rst
```diff
@@ -18,43 +18,170 @@ Sharding
...
+consistency <https://en.wikipedia.org/wiki/ACID#Consistency>`__. Setting
+``r`` or ``w`` to 1 means you will receive a response after only one
+relevant node has responded.
+
+Adding a node
+-------------
+
+To add a node to a cluster, first you must have the additional node
+running somewhere. Make note of the address it binds to, like
+``127.0.0.1``, then ``PUT`` an empty document to the ``/_node``
```
wohali: I'd use a domain name in an actual cluster; running multiple nodes on the same host kind-of defeats the purpose of a cluster and is only really done in dev/demo environments.

src/cluster/sharding.rst
wohali: You could also just point to the documentation over in the Cluster Setup section rather than duplicating this info here. In fact, DRYing this out is probably a good idea...

src/cluster/sharding.rst
```diff
@@ -18,43 +18,170 @@ Sharding
...
+
+Removing a node
+---------------
+
+To remove a node from the cluster, you must first acquire the ``_rev``
+value for the document that signifies its existence:
+
+.. code:: bash
+
+    $ curl "$COUCH_URL:5984/_node/{name}@{address}"
```
wohali: Same problem as above: wrong port, and I'm pretty sure it's `_nodes` not `_node`.

src/cluster/sharding.rst
```diff
@@ -18,43 +18,170 @@ Sharding
...
+    {"_id":"{name}@{address}","_rev":"{rev}"}
+
+Using that ``_rev``, you can delete the node using the ``/_node``
+endpoint:
+
+.. code:: bash
+
+    $ curl -X DELETE "$COUCH_URL:5984/_node/{name}@{address}?rev={rev}"
```
wohali: Same as above.

src/cluster/sharding.rst
```diff
@@ -18,43 +18,170 @@ Sharding
...
+   <div class="alert alert-warning">
+
+**Note**: Before you remove a node, make sure to
+`move its shards <#moving-a-shard>`__
+or else they will be lost.
+
+Moving a shard
+--------------
+
+Moving shards between nodes involves the following steps:
```
wohali: Do we want to talk about just using internal replication as well as the file-copy process? If you are lazy, or don't have remote ssh access, this is easier...though, slower.

src/cluster/sharding.rst
```diff
@@ -18,43 +18,170 @@ Sharding
...
+.. code:: bash
+
+    # one one machine
+    mkdir -p data/shards/{range}
+    # on the other
+    scp $COUCH_PATH/data/shards/{range}/{database}.{timestamp}.couch $OTHER:$COUCH_PATH/data/shards/{range}/
+
+Views are also sharded, and their shards should be moved to save the new
+node the effort of rebuilding the view. View shards live in
+``data/.shards``.
```
wohali: Perhaps include output of an `ls` command here to show actual paths and filenames?

src/cluster/sharding.rst
```diff
@@ -18,43 +18,170 @@ Sharding
...
+
+Updating cluster metadata
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To update the cluster metadata, use the special node-specific ``/_dbs``
+database, accessible via a node's private port, usually at port 5986.
```
wohali: +"This port is only available on the localhost interface for security purposes.

and a note to you: this will hopefully change in 3.0 when port 5986 is dropped completely.

src/cluster/sharding.rst
```diff
@@ -179,123 +328,114 @@ looks like this:
...
-If the shards are large, then you can copy them over manually and only have
-CouchDB sync the changes from the last minutes instead.
+.. code:: bash
 
-.. _cluster/sharding/move:
+    $ curl -X PUT $COUCH_URL:5986/_dbs/{name} -d '{...}'
 
-Moving Shards
-=============
+Replicating from old to new
```
wohali: Nooo no no no. Big -1. Don't do this. We have even talked about disabling the replicator on port 5986 for any purpose other than upgrading from 1.x to 2.x.

Instead just add the shards to the map and CouchDB will manage this thru internal replication itself.

src/cluster/sharding.rst
```diff
@@ -179,123 +328,114 @@ looks like this:
...
-   cluster.
-#. Relax!
+Shard settings for databases can only be set when the database is
+created, precluding live resharding. Instead, to reshard a database, it
+must be regenerated. Here are the steps:
 
-Creating more shards than you need and then move the shards around is called
-presharding. The number of shards you need depends on how much data you are
-going to store. But, creating too many shards increases the complexity without
-any real gain. You might even get lower performance. As an example of this, we
```
wohali: We probably don't want to lose this idea entirely from the docs. I recall from my time at Cloudant creating DBs with very high `q` values under the assumption that new HW was coming soon, and the shards would be re-balanced to the new machines once they were available. I actually forgot we did this until I read this section.

src/cluster/sharding.rst
```diff
@@ -18,43 +18,170 @@ Sharding
...
+That creates a database that is split into 4 shards and 2 replicas,
+yielding 8 shards distributed throughout the cluster.
+
+Quorum
+------
+
+When a CouchDB cluster serves reads and writes, it proxies the request
+to nodes with relevant shards and responds once enough nodes have
+responded to establish
+`quorum <https://en.wikipedia.org/wiki/Quorum_(distributed_computing)>`__.
```
flimzy: I think this should say "to establish a quorum" ?

src/cluster/sharding.rst
```diff
@@ -18,43 +18,170 @@ Sharding
...
+The size of the required quorum can be configured at request time by
+setting the ``r`` parameter for document and view reads, and the ``w``
+parameter for document writes. For example, here is a request that
+specifies that at least two nodes must respond in order to establish
+quorum:
```
flimzy: again "a quorum".

src/cluster/sharding.rst
```diff
@@ -18,43 +18,170 @@ Sharding
...
+
+.. raw:: html
+
+   <div class="alert alert-warning">
+
+**Note**: Before you remove a node, make sure to
+`move its shards <#moving-a-shard>`__
+or else they will be lost.
+
+Moving a shard
```
flimzy: I might suggest putting "Moving a Shard" before "Removing a Node." That's the chronological order one would typically use, and reduces the chance of an accidental removal when someone is just scanning the docs, and further re-enforces the "Note" above.

src/cluster/sharding.rst
```diff
@@ -18,43 +18,170 @@ Sharding
...
+--------------
+
+Moving shards between nodes involves the following steps:
+
+1. Copy the shard file onto the new node.
+2. Update cluster metadata to reflect the move.
+3. Replicate from the old to the new to catch any changes.
+4. Delete the old shard file.
+
```
flimzy: I assume that this process (as well as copying shard files) should be done while the database is offline, or at least read-only, to avoid races that lead to partial copies. Is this a valid assumption?

If so, it might be good to mention that, and how to accomplish this (just stop the couchdb server?) in these steps.

src/cluster/sharding.rst
```diff
@@ -97,27 +224,49 @@ called ``small`` there. Let us look in it. Yes, you can get it with curl too:
...
+shards. The data in the changelog entries and these attributes must
+match. If they do not, the database may become corrupted.
+
+As an example, here is an updated version of the metadata above that
+adds shards to a second node called ``node2``:
+
+.. code:: json
 
     {
         "_id": "small",
```
flimzy: Do you want to rename this `small` to `{name}` for consistency, as well?

src/cluster/sharding.rst
```diff
@@ -97,27 +224,49 @@ called ``small`` there. Let us look in it. Yes, you can get it with curl too:
...
-directory called shards.
+    [
+        "add",
+        "{range}",
+        "{name}@{address}"
+    ]
 
-Use curl to change the ``_dbs/small`` node-local document for small, so it
-looks like this:
+*Note*: You can remove a node by specifying 'remove' instead of 'add'.
```
flimzy: This comment doesn't seem especially relevant here.  Is there a better place, we can put this, where it will be more discoverable by people trying to remove nodes?

src/cluster/sharding.rst
```diff
@@ -179,123 +328,114 @@ looks like this:
...
+This will bring the new shard up to date so that we can safely delete
+the old one.
 
-Moving
-------
+Delete old shard
+~~~~~~~~~~~~~~~~
+
+You can remove the old shard either by deleting its file or by deleting
+it through the private 5986 port:
```
flimzy: Is there a preferred method we should encourage people to use?

I'd vote for through the 5986 port, as it seems less error prone, but this whole process is already pretty manual, so maybe it doesn't really matter unless/until the whole thing can be done through the API.  Maybe others have thoughts? @wohali ?

wohali: I'm ambivalent, with a slight preference for file deletion. The key is to update the cluster metadata first so that the shard is no longer "live" as far as the cluster is concerned; after that, the file is just detritus.

As I mentioned in my review, port 5986 will go away with 3.0, so this section will need at least a partial rewrite sometime before then anyway.

src/cluster/sharding.rst
wohali: On second thought, "presharding" is more related to the concept of CouchDB's sharding in general - we pre-shard databases, meaning we don't re-shard a database after its creation, we decide at creation time where shards go.

Anyway, let's not get into specific recommendations without community consensus. Starting a thread on user@ or dev@ asking for sharding advice based on DB size, doc size, etc. would probably be a good first step.

wohali:
+1 from me but I'd like to see @flimzy and one more core dev to review this, preferably a Cloudanter like @vatamane or similar.

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+5. Clear the target node's maintenance mode.
+6. Update cluster metadata again to remove the source shard(s)
+7. Remove the shard file(s) and secondary index file(s) from the source node.
+
+Copying shard files
+~~~~~~~~~~~~~~~~~~~
+
+Shard files live in the ``data/shards`` directory of your CouchDB
+install. Within those subdirectories are the shard files themselves. For
+instance, for a q=8 database called abc, here is its database shard
```
flimzy: I think "q=8" and "abc" should be double-backtick quoted, for consistency:


    instance, for a ``q=8`` database called ``abc``, here...

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
-.. code-block:: bash
+*q* is the number of database shards to maintain. *n* is the number of
+copies of each document to distribute. The default value for n is 3,
+and for q is 8. With q=8, the database is split into 8 shards. With
+n=3, the cluster distributes three replicas of each shard. Altogether,
+that's 24 shards for a single database. In a default 3-node cluster,
+each node would receive 8 shards. In a 4-node cluster, each node would
+receive 6 shards. We recommend in the general case that the number of
+nodes in your cluster should be a multiple of n, so that shards are
+distributed evenly.
```
flimzy: For consistency with the rest of our documentation, I think we should quote these parameter names, values, and examples.  i.e.:

    The default value for ``n`` is ``3``, and for ``q`` is ``8``.  With ``q=8``, the database ...

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+  data/shards/20000000-3fffffff/abc.1529362187.couch
+  data/shards/40000000-5fffffff/abc.1529362187.couch
+  data/shards/60000000-7fffffff/abc.1529362187.couch
+  data/shards/80000000-9fffffff/abc.1529362187.couch
+  data/shards/a0000000-bfffffff/abc.1529362187.couch
+  data/shards/c0000000-dfffffff/abc.1529362187.couch
+  data/shards/e0000000-ffffffff/abc.1529362187.couch
+
+Since they are just files, you can use ``cp``, ``rsync``,
+``scp`` or other command to copy them from one node to another. For
```
flimzy: other => any other

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+  data/.shards/e0000000-ffffffff/_replicator.1518451591_design
+  data/.shards/e0000000-ffffffff/_replicator.1518451591_design/mrview
+  data/.shards/e0000000-ffffffff/_replicator.1518451591_design/mrview/3e823c2a4383ac0c18d4e574135a5b08.view
+  data/.shards/c0000000-dfffffff
+  data/.shards/c0000000-dfffffff/_replicator.1518451591_design
+  data/.shards/c0000000-dfffffff/_replicator.1518451591_design/mrview
+  data/.shards/c0000000-dfffffff/_replicator.1518451591_design/mrview/3e823c2a4383ac0c18d4e574135a5b08.view
+  ...
+
+.. warning::
```
flimzy: Does it make sense to put this notice at the top of the section?  And should it really be a "Warning" or is "note" (or anything else less scary) appropriate?

As a user, I'd like to know this step is optional (and the cost/benefit of doing or not doing it) before I run through the steps. For a small database, the replication time is likely low enough that I'd rather do the replication, than bother with manually copying files.

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+
+Before we tell CouchDB about the new shards on the node in question, we
+need to set the node to ``true`` maintenance mode. This special mode
+instructs CouchDB to return a ``404 Not Found`` response on the ``/_up``
+endpoint, and ensures it not participate in normal interactive clustered
+requests for its shards. A properly configured load balancer that uses
+``GET /_up`` to check the health of nodes will detect this 404 and
+remove that node from the backend target, preventing any HTTP requests
+from being sent to that node. An example HAProxy configuration to use
+the ``/_up`` endpoint is as follows:
```
flimzy: I suggest rewording this paragraph:

> Before telling CouchDB about the new shards on the node in question, it must be put into maintenance mode. Maintenance mode instructs CouchDB to return a ``404 Not Found`` response on the ``/_up`` endpoint, and ensures it does not participate in normal interactive clustered requests for its shards. A properly configured load balancer that uses ``GET /_up`` to check the health of nodes will detect this ``404`` and remove the node from circulation, preventing requests from being sent to that node. For example, to configure HAProxy to use the ``/_up`` endpoint, use:

flimzy: Incidentally, this doesn't seem like the proper place to describe load balancer configuration. Can we move these details to section 2 ("Installation & First-Time Setup"), and just include a link to it from here?

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+  http-check disable-on-404
+  option httpchk GET /_up
+
+If you do not set maintenance mode, or the load balancer ignores this
+maintenance mode status, after the next step is performed the cluster
+may return incorrect responses when consulting the node in question. You
+don't want this! In the next steps, we will ensure that this shard is
+up-to-date before allowing it to participate in end-user requests.
+
+To set true maintenance mode:
```
flimzy: > To enable maintenance mode:

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+    {"status":"maintenance_mode"}
+
+Finally, check that your load balancer has removed the node from the
+pool of available backend nodes.
+
+Updating cluster metadata to reflect the move
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Now we need to tell CouchDB that the target node (which must already be
+joined to the cluster) should be hosting shard replicas for a given
```
flimzy: Is it worth moving this requirement (that the node is already joined to the cluster) to a separate step? Perhaps a first/preliminary step in the workflow?

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
--------------------------------------------------
+1. Add appropriate changelog entries.
+2. Update the ``by_node`` entries.
+3. Update the ``by_range`` entries.
 
-Start node2 and add it to the cluster. Check in ``/_membership`` that the
-nodes are talking with each other.
+As of this writing, this process must be done manually. **WARNING: Be
+very careful! Mistakes during this process can irreperably corrupt the
+cluster!**
```
flimzy: Can we use the existing warning markup for this, to make it stand out more?

    .. warning::
        Be very careful! Mistakes during this process can irreparably corrupt the cluster!

  As of this writing, this process must be done manually.

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
 
-.. code-block:: javascript
+    ["add", "<range>", "<node-name>"]
+
+The ``<range>`` is the specific shard range for the shard. The ``<node-
+name>`` should match the name and address of the node as displayed in
+``GET /_membership`` on the cluster.
+
+.. warning::
+    When removing a shard from a node, specifying ``remove`` instead of ``add``.
```
flimzy: specifying -> specify

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+write request for a shard on the target node, CouchDB will check if the
+target node's shard(s) are up to date. If it finds they are not up to
+date, it will trigger an internal replication job to complete this task.
+You can observe this happening by triggering a write to the database
+(update a document, or create a new one), while monitoring the
+``/_node/<nodename>/_system`` endpoint, which includes the
+``internal_replication_jobs`` metric.
+
+Once this metric has returned to the baseline from before you wrote the
+document, or is zero (0., it is safe to proceed.
```
flimzy: I think this should be "zero (0)"?  Or maybe we can simplify to:

    document, or is ``0``, it is safe to proceed.

flimzy: Also, can we clarify what "safe to proceed" means?  I guess this just means that the data in the shard is then ready for use?  "safe to proceed" sounds a bit like the user should be careful until this point is reached, but I'm not sure that's really what we mean?

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+
+Clear the target node's maintenance mode
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can now let the node back into the rotation for load balancing by
+putting ``"false"`` to the maintenance mode configuration endpoint, just
+as in step 2.
+
+Verify that the node is in not maintenance mode by performing a ``GET
```
flimzy: Verify that the node is not in maintenance mode ...

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+-----------------------------
 
-.. code-block:: bash
+Database shards can be configured to live solely on specific nodes using
+placement rules.
 
-    mkdir -p data/shards/00000000-7fffffff
+First, each node must be labeled with a zone attribute. This defines
+which zone each node is in. You do this by editing the node’s document
+in the ``/nodes`` database, which is accessed through the “back-door”
```
flimzy: ``/nodes`` -> ``/_nodes``

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+(5986. port. Add a key value pair of the form:
```
flimzy: ... through the "backdoor" port (5986).

flimzy: Is there an official name for this port? It's called "node's private port" above. Can we be consistent in what we call this?  "Local administration port" or similar?

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+placement string.
 
-Reshard? Nope. It cannot be done. So do not create databases with too few
-shards.
+Resharding a database to a new q value
+--------------------------------------
 
-If you can not scale out more because you set the number of shards too low, then
-you need to create a new cluster and migrate over.
+The q value for a database can only be set when the database is created,
```
flimzy: Should quote "q" here, and in the following steps

    ``q``

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+and the number of shard replicas, not every node may have access to
+every shard, but every node knows where all the replicas of each shard
+can be found through CouchDB's internal shard map.
+
+Each request that comes in to a CouchDB cluster is handled by any one
+random coordinating node. This coordinating node proxies the request to
+the other nodes that have the relevant data, which may or may not
+include itself. The coordinating node sends a response to the client
+once a `quorum
+<https://en.wikipedia.org/wiki/Quorum_(distributed_computing)>`__ of
```
willholley: the linked article is perhaps a bit misleading, given the sloppy quorum behaviour (i.e. there is no distributed transaction between the replicas; quorum might fail but data was written anyway).

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+
+    $ curl -X PUT "$COUCH_URL:5984/<doc>?w=2" -d '{...}'
+
+Setting ``r`` or ``w`` to be equal to ``n`` (the number of replicas)
+means you will only receive a response once all nodes with relevant
+shards have responded or timed out, and as such this approach does not
+guarantee `ACIDic consistency
+<https://en.wikipedia.org/wiki/ACID#Consistency>`__. Setting ``r`` or
+``w`` to 1 means you will receive a response after only one relevant
+node has responded.
```
willholley: perhaps a link to https://dx13.co.uk/articles/2015/10/19/couchdb-20s-read-and-write-behaviour-in-a-cluster.html which contains a good description of the read/write behaviour and quorum settings.

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+for databases created after it joined the cluster. To balance the
+distribution of shards from existing databases, they must be moved
+manually.
+
+Moving shards between nodes in a cluster involves the following steps:
+
+1. Copy the shard(s) and any secondary index shard(s) onto the target node.
+2. Set the target node to maintenance mode.
+3. Update cluster metadata to reflect the new target shard(s).
+4. Monitor internal replication to ensure up-to-date shard(s).
```
willholley: also monitor indexing backlog on the new shard

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
-Now, check the node-local ``_dbs_`` database. Here, the metadata for each
-database is stored. As the database is called ``small``, there is a document
-called ``small`` there. Let us look in it. Yes, you can get it with curl too:
+.. code:: bash
 
-.. code-block:: javascript
+    $ curl -X PUT "$COUCH_URL/database-name?q=4&n=2"
 
-    curl -X GET "http://xxx.xxx.xxx.xxx:5986/_dbs/small"
+That creates a database that is split into 4 shards and 2 replicas,
```
willholley: n=2 is probably not an example to promote as you'll get quorum failure if one replica is unavailable. `n=1` is perhaps useful for users who want to maintain CouchDB 1.x consistency semantics and don't need HA.

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+How many shards and replicas each database has can be set at the global
+level, or on a per-database basis. The relevant parameters are *q* and
+*n*.
 
-.. code-block:: bash
+*q* is the number of database shards to maintain. *n* is the number of
+copies of each document to distribute. The default value for n is 3,
+and for q is 8. With q=8, the database is split into 8 shards. With
+n=3, the cluster distributes three replicas of each shard. Altogether,
+that's 24 shards for a single database. In a default 3-node cluster,
```
mikerhodes: "24 shard replicas for a single database". There's still eight shards :)

src/cluster/sharding.rst
mikerhodes: Agreed, I'd use n=6 or similar to encourage any diffs from 3 to be positive in terms of data protection.

src/cluster/sharding.rst
mikerhodes: I'd agree that this config note should be at least in a "pre-requisites" section above the first copy step -- if I were moving shards around it'd be frustrating if I found in the middle I needed to re-configure my load balancer. (On the other hand, if I hadn't read through all the steps beforehand, I'm kind of asking for this kind of surprise).

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+
+Now we need to tell CouchDB that the target node (which must already be
+joined to the cluster) should be hosting shard replicas for a given
+database.
+
+To update the cluster metadata, use the special ``/_dbs`` database,
+which is an internal CouchDB database that maps databases to shards and
+nodes. It is accessible only via a node's private port, usually at port
+5986. By default, this port is only available on the localhost interface
+for security purposes.
```
mikerhodes: IIRC this only needs performing on one node in the cluster as the `_dbs` database is replicated internally -- definitely worth mentioning this!

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
-After PUTting this document, it's like magic: the shards are now on node2 too!
-We now have ``n=2``!
+Now you can ``PUT`` this new metadata:
 
-If the shards are large, then you can copy them over manually and only have
-CouchDB sync the changes from the last minutes instead.
+.. code:: bash
 
-.. _cluster/sharding/move:
+    $ curl -X PUT $COUCH_URL:5986/_dbs/{name} -d '{...}'
```
mikerhodes: Suggest here sticking to use of `localhost` as above.

src/cluster/sharding.rst
mikerhodes: Yes, something that segues to the next point, like "is 0, the shard replica is ready to serve data and we can bring the node out of maintenance mode".

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+
+Once this metric has returned to the baseline from before you wrote the
+document, or is zero (0., it is safe to proceed.
+
+Clear the target node's maintenance mode
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can now let the node back into the rotation for load balancing by
+putting ``"false"`` to the maintenance mode configuration endpoint, just
+as in step 2.
```
mikerhodes: I think this is a little misleading -- the reason the node is in mm is because it shouldn't be servicing requests for the data it's hosting, because it had an out of date shard replica of the shard we're moving. That it was out of the LB's rotation is a side-effect rather than intended effect in this particular task.

So I think this could be something like "You can now let the node start servicing data requests by putting...".

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+Update cluster metadata again to remove the source shard
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Add, then delete
-----------------
+Now, remove the source shard from the shard map the same way that you
+added the new target shard to the shard map in step 2. Be sure to add
+the ``["remove", <range>, <source-shard>]`` entry to the end of the
+changelog as well as modifying both the by_node and by_range sections of
+the database metadata document.
```
mikerhodes: The backticks on `by_node` and `by_range`/

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
-you might be caught by the following mistake:
+Remove the shard and secondary index files from the source node
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-#. Copy the shard onto a new node.
-#. Update the metadata to use the new node.
-#. Delete the shard on the old node.
-#. Oh, no!: You have lost all writes made between 1 and 2.
+Finally, you can remove the source shard by deleting its file from the
+command line on the source host, along with any view shards:
```
mikerhodes: "remove the source shard replica", "view shard replicas"?

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+.. code::bash
 
-Moving
-------
+    $ rm <couch-dir>/data/shards/<range>/<dbname>.<datecode>.couch
+    $ rm -r <couch-dir>/data/.shards/<range>/<dbname>.<datecode>*
 
-When you get to ``n=3`` you should start moving the shards instead of adding
-more replicas.
+Congratulations! You have moved a database shard. By adding and removing
```
mikerhodes: "shard replica".

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+also known as a shard map.
 
-We will stop on ``n=2`` to keep things simple. Start node number 3 and add it to
-the cluster. Then create the directories for the shard on node3:
+Specifying database placement
+-----------------------------
 
-.. code-block:: bash
+Database shards can be configured to live solely on specific nodes using
+placement rules.
```
mikerhodes: Zones are more about initial placement when the database is created. Given we've just been talking about manually moving shards, it's worth calling out that we've changed gear a bit.

"You can configure CouchDB to put shard replicas on certain nodes at database create time using placement rules."

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+1. Create a temporary database with the desired shard settings, by
+   specifying the q value as a query parameter during the PUT
+   operation.
+2. Replicate the primary database to the temporary one. Multiple
+   replications may be required if the primary database is under
+   active use.
+3. Delete the primary database. **Make sure nobody is using it!**
+4. Recreate the primary database with the desired shard settings.
+5. Replicate the temporary back to the primary.
+6. Delete the temporary database.
```
mikerhodes: This more describes an "offline" resharding. Rather than having the note in step 3, to be safe, I think there should be extra instructions:

1. Prior to (2), "stop clients accessing the database".
2. After (5), "clients can now access the database again".

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+include itself. The coordinating node sends a response to the client
+once a `quorum
+<https://en.wikipedia.org/wiki/Quorum_(distributed_computing)>`__ of
+database nodes have responded; 2, by default.
+
+The size of the required quorum can be configured at
+request time by setting the ``r`` parameter for document and view
+reads, and the ``w`` parameter for document writes. For example, here
+is a request that specifies that at least two nodes must respond in
+order to establish a quorum:
```
mikerhodes: I think it's very important here to note that `r` and `w` don't change the behaviour of the database _at all_ excepting the coordinator's behaviour. Right now this is still a bit misleading on the roles of `r` and `w`.

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+``w`` to 1 means you will receive a response after only one relevant
+node has responded.
+
+.. _cluster/sharding/move:
+
+Moving a shard
+--------------
+
+This section describes how to manually place and replace shards, and how
+to set up placement rules to assign shards to specific nodes. These
```
mikerhodes: I think it's actually the next section that talks about placement rules.

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+means you will only receive a response once all nodes with relevant
+shards have responded or timed out, and as such this approach does not
+guarantee `ACIDic consistency
+<https://en.wikipedia.org/wiki/ACID#Consistency>`__. Setting ``r`` or
+``w`` to 1 means you will receive a response after only one relevant
+node has responded.
+
+.. _cluster/sharding/move:
+
+Moving a shard
```
mikerhodes: This seems like a nice write up of this fairly complicated process :)

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+CouchDB nodes have a ``etc/default.ini`` file with a section named
+``[cluster]`` which looks like this:
 
-If you look in the directory ``data/shards`` you will find the 2 shards.
+::
 
-.. code-block:: text
+    [cluster]
+    q=8
+    n=3
```
mikerhodes: Is it worth noting that `n=3` is a generally accepted value for high data resilience to node loss? Unfortunately I don't have a link to something explaining why :(

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
 .. _cluster/sharding:
 
-========
-Sharding
-========
+================
+Shard Management
+================
 
 .. _cluster/sharding/scaling-out:
```
mikerhodes: Is it worth adding a bit of an intro about what this document contains?

src/cluster/sharding.rst
wohali: Our documentation should be as self-contained as possible, and definitely not rely on 3rd party resources to explain what we refuse to write up ourselves... ;)

wohali: My point being - if the explanation is useful, email the original author and ask for permission to include it in the docs!

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
-Normally you start small and grow over time. In the beginning you might do just
-fine with one node, but as your data and number of clients grows, you need to
-scale out.
+A `shard
+<https://en.wikipedia.org/wiki/Shard_(database_architecture)>`__ is a
+horizontal partition of data in a database. Partitioning data into
+shards and distributing copies of each shard (called "shard replicas" or
+just "replicas") to different nodes in a cluster gives the data greater
+durability against node loss. CouchDB clusters automatically shard
+databases and distribute the subsections of documents that compose each
```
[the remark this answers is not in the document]
garbados: You're right, it should read "subset" of documents. I'll update it accordingly.

src/cluster/sharding.rst
garbados: That makes sense. I'll make the change.

src/cluster/sharding.rst
wohali: "node-local port" and "clustered port" are the most common terms.

src/cluster/sharding.rst
garbados: Could you clarify? What metrics should one observe to track this?

garbados:
I added a new commit that responds to a lot of the feedback folks left. I'll respond to the outstanding comments that remain so I can figure out any additional changes it needs.

src/cluster/sharding.rst
garbados: I added something to this effect:

> This document discusses how sharding works in CouchDB along with how to
> safely add, move, remove, and create placement rules for shards and
> shard replicas.

src/cluster/sharding.rst
garbados: Thanks!

src/cluster/sharding.rst
```diff
@@ -12,290 +12,498 @@
...
+  data/.shards/c0000000-dfffffff
+  data/.shards/c0000000-dfffffff/_replicator.1518451591_design
+  data/.shards/c0000000-dfffffff/_replicator.1518451591_design/mrview
+  data/.shards/c0000000-dfffffff/_replicator.1518451591_design/mrview/3e823c2a4383ac0c18d4e574135a5b08.view
+  ...
+
+Set the target node to ``true`` maintenance mode
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Before telling CouchDB about the new shards on the node in question, it
```
janl: it is ambiguous whether the `it` refers to CouchDB or the node in question. This could be read as disabling the cluster, instead of just the new node.

src/cluster/sharding.rst
janl: Node management is not part of this doc, and is covered here: https://github.com/apache/couchdb-documentation/blob/master/src/cluster/nodes.rst if we don’t have a link above, it should be added

src/cluster/sharding.rst
```diff
@@ -12,290 +12,498 @@
...
+--------------
+
+This section describes how to manually place and replace shards. These
+activities are critical steps when you determine your cluster is too big
+or too small, and want to resize it successfully, or you have noticed
+from server metrics that database/shard layout is non-optimal and you
+have some "hot spots" that need resolving.
+
+Consider a three-node cluster with q=8 and n=3. Each database has 24
+shards, distributed across the three nodes. If you add a fourth node to
```
janl: link `add a fourth node` to https://github.com/apache/couchdb-documentation/blob/master/src/cluster/nodes.rst

src/cluster/sharding.rst
```diff
@@ -12,290 +12,498 @@
...
+target node's shard(s) are up to date. If it finds they are not up to
+date, it will trigger an internal replication job to complete this task.
+You can observe this happening by triggering a write to the database
+(update a document, or create a new one), while monitoring the
+``/_node/<nodename>/_system`` endpoint, which includes the
+``internal_replication_jobs`` metric.
+
+Once this metric has returned to the baseline from before you wrote the
+document, or is ``0``, the shard replica is ready to serve data and we
+can bring the node out of maintenance mode.
```
janl: Nice!

src/cluster/sharding.rst
```diff
@@ -12,290 +12,498 @@
...
+
+Clear the target node's maintenance mode
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can now let the node start servicing data requests by
+putting ``"false"`` to the maintenance mode configuration endpoint, just
```
janl: this is a valid use of `just`, of course ;)

src/cluster/sharding.rst
```diff
@@ -12,290 +12,498 @@
...
+  data/shards/00000000-1fffffff/abc.1529362187.couch
+  data/shards/20000000-3fffffff/abc.1529362187.couch
+  data/shards/40000000-5fffffff/abc.1529362187.couch
+  data/shards/60000000-7fffffff/abc.1529362187.couch
+  data/shards/80000000-9fffffff/abc.1529362187.couch
+  data/shards/a0000000-bfffffff/abc.1529362187.couch
+  data/shards/c0000000-dfffffff/abc.1529362187.couch
+  data/shards/e0000000-ffffffff/abc.1529362187.couch
+
+Since they are just files, you can use ``cp``, ``rsync``,
```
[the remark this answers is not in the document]
garbados: This is an important point. I'll prune the document for such instances.

src/cluster/sharding.rst
```diff
@@ -12,290 +12,490 @@
...
+
+Moving shards between nodes in a cluster involves the following steps:
+
+1. Copy the shard(s) and any secondary index shard(s) onto the target node.
+2. Set the target node to maintenance mode.
+3. Update cluster metadata to reflect the new target shard(s).
+4. Monitor internal replication to ensure up-to-date shard(s).
+5. Clear the target node's maintenance mode.
+6. Update cluster metadata again to remove the source shard(s)
+7. Remove the shard file(s) and secondary index file(s) from the source node.
```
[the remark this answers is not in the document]
garbados: I'd really like to do this but I can't figure out the multi-line formatting for links, as the links exceed the line character limit unless they are split across multiple lines, but splitting them breaks formatting. I am not sure how to resolve the issue.

garbados:
I added another couple commits to respond to feedback. Currently the build fails because, in an effort to turn the steps for moving a shard into links to each section, I broke formatting and I don't know how to fix it. [These lines](https://github.com/apache/couchdb-documentation/pull/268/commits/bad9ec19393899a3335b37ae948069b49340dfb0#diff-c4a20adee869069542e57a24c680c5c3R141) are too long when formatted as a reference, but splitting it across multiple lines breaks the reference. I am not sure how to resolve this issue.

janl:
I couldn’t find anything in terms of multiline `:ref:` labels either :/

I pushed a commit that does some trickery, so not the full line is referenced, to unblock this PR.

PR #270, opened 2018-04-09
Update _stats and _system sections

### LINKED ISSUE #235: /_stats -> /_node/<nodename>/_stats or /_node/_local/_stats
>The current docs for 2.0 onwards show that it is possible to get to the _stats via /_stats but it actually requires /_node//_stats

As of 2.1.1 we also now support `/_node/_local/_stats` which was documented in the [release notes only](http://docs.couchdb.org/en/2.1.1/whatsnew/2.1.html#version-2-1-1).

wohali:
Closes #235.

### DIFF
--- a/src/api/server/common.rst
+++ b/src/api/server/common.rst
@@ -1315,18 +1315,22 @@ error.
 
 .. _api/server/stats:
 
-===========
-``/_stats``
-===========
+=============================
+``/_node/{node-name}/_stats``
+=============================
 
-.. http:get:: /_stats
+.. http:get:: /_node/{node-name}/_stats
     :synopsis: Returns server statistics
 
     The ``_stats`` resource returns a JSON object containing the statistics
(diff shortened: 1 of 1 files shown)


src/api/server/common.rst
```diff
@@ -1354,150 +1358,139 @@ error.
...
 
 .. code-block:: http
 
-    GET /_stats/couchdb/request_time HTTP/1.1
+    GET /_node/_local/_stats/couchdb/request_time HTTP/1.1
 
 This returns an entire statistics object, as with the full request, but
-containing only the request individual statistic. Hence, the returned structure
-is as follows:
+containing only the request individual statistic.
```
flimzy: I know you didn't add this wording, but "request individual statistic" is a bit confusing to me. Is there a way to clarify that?  I'm not even sure what it means, so I don't have a suggestion to offer myself.

PR #271, opened 2018-04-12
Bump curl earlier in intro, fix for Windows

wohali:
cURL should be explained before our first example that uses it.

On Windows, special escaping is required; explain the syntax necessary for beginners right at the start.

### DIFF
--- a/src/intro/curl.rst
+++ b/src/intro/curl.rst
@@ -51,6 +51,26 @@ clarity):
 
         shell> curl 'http://couchdb:5984/_uuids?count=5'
 
+.. lint: ignore errors for the next 15 lines
+
+.. note::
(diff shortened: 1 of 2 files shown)


src/intro/curl.rst
```diff
@@ -51,6 +51,26 @@ clarity):
 
         shell> curl 'http://couchdb:5984/_uuids?count=5'
 
+.. lint: ignore errors for the next 15 lines
+
+.. note::
+    On Microsoft Windows, use double-quotes anywhere you see single-quotes in
+    the following examples. and double double-quotes ("") anywhere you see
```
flimzy: The period (.) after "example" should be a comma (,).

PR #272, opened 2018-04-12
Add nginx docs

wohali:
Adds basic docs on using nginx as a reverse proxy.

Includes a nice header saying we still thing HAProxy is better.

This is part of the wiki migration project.

### DIFF
--- a/src/best-practices/nginx.rst
+++ b/src/best-practices/nginx.rst
@@ -1,0 +1,125 @@
+.. Licensed under the Apache License, Version 2.0 (the "License"); you may not
+.. use this file except in compliance with the License. You may obtain a copy of
+.. the License at
+..
+..   http://www.apache.org/licenses/LICENSE-2.0
+..
+.. Unless required by applicable law or agreed to in writing, software
+.. distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+.. WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+.. License for the specific language governing permissions and limitations under
+.. the License.
+
+.. _best-practices/nginx:
+
+========================
+nginx as a Reverse Proxy
+========================
+
+CouchDB recommends the use of `HAProxy`_ as a load balancer and reverse proxy.
+The team's experience with using it in production has shown it to be superior
+for configuration and montioring capabilities, as well as overall performance.
+
+CouchDB's sample haproxy configuration is present in the `code repository`_ and
+release tarball as ``rel/haproxy.cfg``.
+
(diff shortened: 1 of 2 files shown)


src/best-practices/nginx.rst
```diff
@@ -0,0 +1,134 @@
...
+.. code-block:: text
+
+    location / {
+        proxy_pass http://localhost:5984;
+        proxy_redirect off;
+        proxy_set_header Host $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    }
+
+Continuous Replication through nginx
```
flimzy: I don't think "Replication" needs to be capitalized to be more consistent

src/best-practices/nginx.rst
```diff
@@ -0,0 +1,134 @@
...
+====================================
+
+The basic configuration above will **not** work correctly when attempting
+continuous replication. To rectify this problem, use the following block:
```
flimzy: Is it worth explaining why this different block is necessary?  Or just simplify the docs, and only show the "fully functional" version?

src/best-practices/nginx.rst
```diff
@@ -0,0 +1,134 @@
...
+        proxy_buffering off;
+        proxy_set_header Host $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    }
+
+Reverse proxying CouchDB in a subdirectory with nginx
+=====================================================
+
+It is useful to provide CouchDB as a subdirectory of your overall domain,
+especially to avoid CORS concerns. Here's an excerpt of a basic nginx
```
flimzy: Would "It can be useful..." be more appropriate?

src/best-practices/nginx.rst
```diff
@@ -0,0 +1,134 @@
...
+    location /couchdb {
+        rewrite /couchdb/(.*) /$1 break;
+        proxy_pass http://localhost:5984;
+        proxy_redirect off;
+        proxy_set_header Host $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    }
+
+Note that in the above configuration, the *Verify Installation* link in
+Fauxton may not succeed.
```
flimzy: May not, or will not?

src/best-practices/nginx.rst
wohali: +1, will merge. Thought about that at the time, then forgot to actually do it.

src/best-practices/nginx.rst
wohali: Unknown, and I don't have the time to set this up to test - so it's may.

PR #274, opened 2018-05-11
Callout the 4GB -> 64MB limit change

ahayes:
## Overview

Ran into an issue upgrading 1.6 to 2.1.1 in-place using couchup. The 1.6 DB had large design documents with many attachments and the upgrade was not successful as a result of a change in maximum size checking and a much lower default value. Our application was also affected as it attempted to make large requests. The upgrade notes highlighted the new parameter but did not describe the decrease in default value or how to identify an issue. The 413 HTTP code as a symptom of hitting the limit was only mentioned in the reference documentation for the parameter.

## Checklist


ahayes:
Thanks @rnewson for endless patience while helping me hunt this one down!

ahayes:
Can this be back-ported to 2.1.1 documentation?

PR #276, opened 2018-05-16
Remove mentions on external process shenanigans

janl:
This never really made it to 2.x and won’t be a thing going forward. At the moment it only confuses folks by being documented.


wohali:
We need to remove all of this from couch as well:

```
default.ini:os_daemons={couch_os_daemons, start_link, []}
default.ini:; [os_daemons]
default.ini:; [os_daemon_settings]
local.ini:[os_daemons]
local.ini:[update_notification]

```

wohali:
You have a +1, merge away

PR #287, opened 2018-07-10
Drop unsupported node query server, plugins from 2.x

### LINKED ISSUE #286: NodeJS Query Server, missing documentation
http://docs.couchdb.org/en/2.1.1/experimental.html?highlight=query%20server#nodejs-query-server
This is reported not to work.

If anyone have any documentation for how to use this, I would be interested in testing it.
Nodejs server via proxy can be used in combination with rewrite as a function, when this bug is fixed https://github.com/apache/couchdb/issues/1407

wohali:
Neither of these features made it into 2.x and should have been scrubbed from the docs a while ago.

Fixes #286.

### DIFF
--- a/src/experimental.rst
+++ b/src/experimental.rst
@@ -18,69 +18,17 @@ Experimental Features
 
 This is a list of experimental features in CouchDB. They are included in
 a release because the development team is requesting feedback from the
-larger developer community. As such, please play around with these features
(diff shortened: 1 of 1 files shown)


wohali:
Note the plan is now to replace the nodejs query server with couch-chakra, see apache/couchdb#1334.

PR #293, opened 2018-07-13
fix proxy auth docs, goes with https://github.com/apache/couchdb/pull/1436

janl:


src/api/server/authn.rst
```diff
@@ -272,15 +272,10 @@ To obtain the first token and thus authenticate a user for the first time, the
 Proxy Authentication
 ====================
 
-.. note::
-    To use this authentication method make sure that the
-    ``{couch_httpd_auth, proxy_authentication_handler}`` value in added to the
-    list of the active :config:option:`chttpd/authentication_handlers`:
-
```
wohali: I would restore this block now that you're leaving the block below it, and `s/couch_httpd/chttpd`.

PR #297, opened 2018-07-18
Configure cluster Erlang port range using vm.args

d3alek:
Ports specified in sys.config as per previous documentation seem to be ignored by CouchDB

## Overview

Current documentation suggests putting the distributed Erlang port range restriction in `sys.config`, however in my experience this does not get picked up by CouchDB. What works for me is specifying the port range in `vm.args` instead.  I am using a recent build from source (`62f71c0b0`).

I am aware I may not be putting `sys.config` in the correct place to be picked up by CouchDB - if that's the case and modifying `sys.config` works for you then let me know and I will change the PR to simply clarify the correct location for `sys.config`. 

## Checklist

- [x ] Documentation is written and is accurate;
- [x ] `make check` passes with no errors

### DIFF
--- a/src/cluster/setup.rst
+++ b/src/cluster/setup.rst
@@ -119,31 +119,16 @@ To close the shells, run in both:
 Make CouchDB use the open ports.
 --------------------------------
 
-Open ``sys.config``, on all nodes, and add ``inet_dist_listen_min, 9100`` and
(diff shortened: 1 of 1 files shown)


wohali:
Thank you for the important fix, @d3alek ! And welcome to the CouchDB contributor community! :tada:

PR #300, opened 2018-07-24
2.2.0 release notes, increase max line length to 90

wohali:
This is now ready to go except for adding missing Fauxton/documentation release notes, which we can leave out if we want to.

I feel the bump in the linter from 80 to 90 columns is justified based on increased support for this in the wider Python community at this point, but don't need it, and can rework if anyone finds this offensive. (Remember, this is a self-imposed style; RST itself doesn't have any line limits.)

### DIFF
--- a/ext/linter.py
+++ b/ext/linter.py
@@ -196,8 +196,8 @@ def whitespace_committee(file):
 
 @register_rule
 def terminal_emulator(file):
-    """Terminal emulator has screen limited by 80 chars wide, so it ensures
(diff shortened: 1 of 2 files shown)


ext/linter.py
```diff
@@ -196,8 +196,8 @@ def whitespace_committee(file):
 
 @register_rule
 def terminal_emulator(file):
```
kxepal: I guess, function name should be changed as well (;

PR #312, opened 2018-08-08
Disclose CVE 2018-11769

wohali:


wohali:
3 +1s from @rnewson @janl @garrensmith via CouchDB security IRC channel.

PR #314, opened 2018-08-10
Fix version for rewrites as a function

BigBlueHat:
Follow-up to https://github.com/apache/couchdb/issues/1410

I'm not sure which specific version of CouchDB rewrites-as-a-function was introduced, but it was sometime after CouchDB 2.0.0--hence the `2.x` moniker.

Updated it if you need more accuracy. 😄 

Cheers!
🎩


wohali:
Here's where the code landed:

https://github.com/apache/couchdb/pull/370

That would have landed in 2.0.0.

PR #318, opened 2018-08-14
Add Caddy Server reverse-proxy config examples incl. cluster load balancing

detroitenglish:
## Overview

This PR adds configuration examples for [Caddy](https://github.com/mholt/caddy) web-server as a reverse-proxy in front of CouchDB.

In particular, these examples include a basic configuration for applying Caddy's builtin load-balancing and health-check features for reverse-proxying to a CouchDB cluster. The nginx example lacks this, as load balancing + health check are paid-version-only features.

Lastly I believe Caddy's automatic SSL/TLS will help to alleviate a major pain point for developers using reverse-proxies: certificate management.

(I've used Caddy for reverse-proxy-to-cluster well over a year now; performance has been top-notch)

## Testing recommendations
N/A

## GitHub issue number
N/A

## Related Pull Requests
N/A

## Checklist


src/best-practices/caddy.rst
```diff
@@ -0,0 +1,167 @@
...
+
+CouchDB recommends the use of `HAProxy`_ as a load balancer and reverse proxy.
+The team's experience with using it in production has shown it to be superior
+for configuration and monitoring capabilities, as well as overall performance.
+
+CouchDB's sample haproxy configuration is present in the `code repository`_ and
+release tarball as ``rel/haproxy.cfg``.
+
+However, ``Caddy`` is a suitable alternative. Below are instructions on
+configuring Caddy appropriately.
```
wohali: Since you're copying over the content from `src/best-practices/nginx.rst`, why not rename that file to `src/best-practices/reverse-proxy.rst` and just add Caddy below the Nginx section? I don't like the idea of having to maintain both of these files in the doc tree in case the haproxy.cfg file changes, or our feelings about HAProxy itself changes.

src/best-practices/caddy.rst
```diff
@@ -0,0 +1,167 @@
...
+            transparent
+        }
+
+    }
+
+Note that, because Caddy is https-by-default, you must explicitly include the
+``http://`` protocol in the site address if you do NOT want Caddy
+to automatically acquire and install an SSL certificate and begin accepting
+``https`` connections on port 443.
+
```
wohali: Before you get into subdirectory reverse proxying, are there any issues with special configuration for replication through Caddy as there are for nginx?

Specifically, the `Host` and `X-Forwarded-For` headers are mandatory to solve problems with CouchDB behind a reverse proxy properly.

Have you tested this configuration with both push and pull replication to a remote CouchDB somewhere else?

src/best-practices/caddy.rst
```diff
@@ -0,0 +1,167 @@
...
+======================================================
+
+Here's a basic excerpt from a Caddyfile in
+``/<path>/<to>/<site>/Caddyfile``. This will proxy and evenly distribute all
+requests from ``http(s)://domain.com/...`` among 3 CouchDB cluster nodes
+at ``localhost:15984``, ``localhost:25984`` and ``localhost:35984``.
+
+Caddy will check the status, i.e. health, of each node every 5 seconds;
+if a node goes down, Caddy will avoid proxying requests to that node until it
+comes back online.
```
wohali: Does Caddy's health check properly respond to non-200 responses on the `health_check` endpoint? There are important scenarios in shard rebalancing where a node is placed into a state where it is still online, but not participating in clustered responses, where it will return a 4xx response code. Caddy needs to understand these and stop redirecting requests to that node until a `200` response is seen again.

src/best-practices/caddy.rst
```diff
@@ -0,0 +1,167 @@
...
+    domain.com {
+
+        import /path/to/other/config.caddy # logging, error handling etc.
+
+        proxy / http://localhost:15984 http://localhost:25984 http://localhost:35984 {
+            policy round_robin
+            health_check /_up
+            health_check_duration 5s
+            try_interval 500ms
+            timeout 1h
```
wohali: Does this mean that after seeing a successful `200`, Caddy won't check again for another hour? That could be catastrophic in a cluster setting where nodes go offline.

src/best-practices/caddy.rst
```diff
@@ -0,0 +1,167 @@
...
+            transparent
+        }
+
+    }
+
+Note that in the above configuration, the *Verify Installation* link in
+Fauxton may not succeed.
```
wohali: Can you provide more details? Can you test this in 2.2.0? We now support Fauxton in a subdirectory, so this warning over in Nginx land is probably obsolete with this release.

src/best-practices/caddy.rst
```diff
@@ -0,0 +1,167 @@
...
+
+    domain.com {
+
+        import /path/to/other/config.caddy # logging, error handling etc.
+
+        basicauth /couchdb couch_username couchdb_password
+
+        proxy /couchdb localhost:5984 {
+            transparent
+            header_upstream -Authorization
```
wohali: Is an example available that correctly passes the `X-Auth-CouchDB-*` headers downstream? I didn't write one for Nginx, but arguably there should be one in there at some point. See http://docs.couchdb.org/en/stable/api/server/authn.html#proxy-authentication for details.

src/best-practices/caddy.rst
```diff
@@ -0,0 +1,167 @@
...
+
+.. code-block:: text
+
+    domain.com {
+
+        import /path/to/other/config.caddy # logging, error handling etc.
+
+        proxy / localhost:5984 {
+            transparent
+            header_upstream x-forwarded-ssl on
```
wohali: What about the `X-Forwarded-For` and `Host` headers, are these automatic? Same comment as earlier in this file. Replications will likely fail without these set correctly.

src/best-practices/caddy.rst
detroitenglish: From the [Caddy http.proxy documentation](https://caddyserver.com/docs/proxy):

> If a backend returns a status code of 200-399, then that backend is considered healthy.

src/best-practices/caddy.rst
detroitenglish: `transparent` is a preset which includes all the necessary headers for CouchDB. Clarified with a note under Basic Config: https://github.com/apache/couchdb-documentation/pull/318/commits/dfaa2c6b286e036f8a8677447138a4c0dfb1ee6c

In short, it's:
```
header_upstream Host {host}
header_upstream X-Real-IP {remote}
header_upstream X-Forwarded-For {remote}
header_upstream X-Forwarded-Proto {scheme}
```

I've used Caddy for 3+ node clusters for a couple years now, and it's handled all the replications I've thrown at it without issue

src/best-practices/caddy.rst
detroitenglish: Caddy does not add, remove or alter up- or downstream headers unless explicitly instructed to do so. So in this case it didn't seem necessary as it's default behavior.

src/best-practices/caddy.rst
detroitenglish: Clarified with a note under Basic Config: https://github.com/apache/couchdb-documentation/pull/318/commits/dfaa2c6b286e036f8a8677447138a4c0dfb1ee6c

src/best-practices/caddy.rst
detroitenglish: Nice idea! Merged nginx and Caddy into section 'Reverse Proxies' - https://github.com/apache/couchdb-documentation/pull/318/commits/a7bd43ad8d7a4fe2adb2f0c7586203bdf7562f24

src/best-practices/caddy.rst
detroitenglish: Also removed from nginx section - https://github.com/apache/couchdb-documentation/pull/318/commits/8b6d295aca4f4444de99a00617f7c1b5386cb879

src/best-practices/reverse-proxies.rst
```diff
@@ -0,0 +1,269 @@
...
+instead, e.g. ``{$COUCH_PW}``.
+
+This setup leans entirely on Caddy performing authorization, and forwarding
+requests to CouchDB with no authentication (with CouchDB in Admin Party mode).
+For a better solution, see :ref:`api/auth/proxy`.
+
+SSL/TLS with Caddy
+------------------
+
+Caddy is http-by-default, and will automatically acquire, install, activate and,
```
flimzy: Did you mean "https-by-default"?

detroitenglish: Good catch - thanks!

src/best-practices/caddy.rst
wohali: Turns out that this is actually not a good option performance-wise for CouchDB right now - and yes, we know this isn't ideal.

This is why we do [this](https://github.com/apache/couchdb/blob/master/rel/haproxy.cfg#L27) in the haproxy config.

Would you be willing to test changing this on one of your clusters?

detroitenglish: Confession: I've always had Caddy handle timeouts... 🙈 I use the default setting of 30s

The PR-example in question was really my attempt to comply with the existing haproxy config 'spec'; obviously I mistranslated* the `timeout server 3600000` in haproxy. 

So the haproxy config's `timeout client 150000`  i.e. 2.5 minutes setting is the recommended timeout duration?

> Turns out that this is actually not a good option performance-wise for CouchDB right now

\* In my defense, yeah, this ^ is kind weird :)

------------

Below is the _actual_ config I'm running in prod commented with definitions and some defaults included for added clarity.

```
proxy /db 127.0.0.1:5984 {$REMOTE_COUCH_ONE} {$REMOTE_COUCH_TWO} {

  # My custom LB settings [Caddy default setting in brackets]
  policy least_conn # Select backend with fewest active connections [random]
  fail_timeout 30s  # How long to remember a failed request to a backend [0s]
  max_fails 2       # Failed reqs within fail_timeout to consider a backend down [1]
  try_duration 10s  # How long client may hang as proxy looks for a backend [0s]

  # Caddy default LB settings (included for reference)
  timeout 30s         # duration before timing out the connection to upstream
  try_interval 250ms  # how long to wait between selecting another upstream host

  # Health check settings
  health_check /_up         # ...
  health_check_interval 10s # ...
  health_check_timeout 15s  # Deadline to respond, otherwise req is a failure

  # Header settings for requests to CouchDB
  transparent
  header_upstream X-Forwarded-Ssl on
  header_upstream -Authorization

  # Header settings for response back to client
  header_downstream X-Lol-Its {$API_INSTANCE_ID}  # ID for load testing
  header_downstream -Server                       # Remove server header

  without /db
}
```
Here's the [documentation on Caddy's http.proxy](https://caddyserver.com/docs/proxy#syntax) directive, for further reference.  

It'd be great if we can knock out a high-performance example together - let me know what you think!

wohali: So the issue isn't specifically the timeouts, it's that HAProxy currently closes out every request on the backend (i.e. HAProxy <-> CouchDB) when a request finishes, so there really isn't a reuseable connection pool at the moment.

I've personally seen some weird behaviour in production when people have used LB-managed connection pools under heavy load that cleared up when we disabled the connection pooling. I know for a fact it's especially bad if you use SSL between the reverse proxy (in the case I'm thinking of, it was an F5 BigIP) and CouchDB, but that has more to do with the poor implementation by Erlang of SSL. (We put a workaround into 2.2.0 for this via a `vm.args` config setting to auto-close any SSL connections after 300s.)

So the only change I'm recommending to the docs here is not to recommend backend connection pooling, when we recommend the exact opposite in `haproxy.cfg`. You could choose to leave out that part of the configuration, as I did with the Nginx config (the point was to provide a working example, not a tuned high-perf one), or we can hold up this PR until I have time to build a test harness and compare pooled vs. unpooled, which might be months.

My recommendation: remove the `timeout` line entirely. The Caddy default is 30s, which is fine, and if people want something higher, they can do that.

If you want to match the haproxy config more exactly, I would add `keepalive 0`.

detroitenglish: > I know for a fact it's especially bad if you use SSL between the reverse proxy... and CouchDB

Saw this as well, so I offload encryption to `stunnel` and trick the Couches into thinking all their cluster buddies are on loopback.

It looks like WireGuard will be pulled into the Linux kernel soon - how cool would it be if cluster setup bootstrapped its own VPN in the background...!

>My recommendation: remove the timeout line entirely. The Caddy default is 30s, which is fine, and if people want something higher, they can do that.
>
>If you want to match the haproxy config more exactly, I would add keepalive 0.

Implemented - https://github.com/apache/couchdb-documentation/pull/318/commits/74df1dc7a5dff3ecf35bd886b08daa054d1cc457  👍 Thanks for your detailed feedback, as always!

src/best-practices/caddy.rst
detroitenglish: > If you specifically want to "login" to CouchDB as a user when set up in proxy auth mode, you specify `X-Auth-CouchDB-UserName: wohali` and possibly also e.g. `X-Auth-CouchDB-Roles: admin,writer`.

I got the impression somewhere that `proxy_authentication_handler` is on its way out... is that not the case?

wohali:
@detroitenglish Thanks again for this contribution - very useful!

src/best-practices/caddy.rst
wohali: No, CouchDB proxy (where CouchDB is the reverse proxy for another service) is gone. CouchDB auth proxy will stay for the foreseeable future...

PR #320, opened 2018-08-16
correct the note about replicator/auth_plugins

### LINKED ISSUE #319: Documentation: replicator auth_plugin is actually enabled by default
## Current Behavior
[Documentation](http://docs.couchdb.org/en/2.2.0/config/replicator.html) says: `In version 2.2, the session plugin is considered experimental and is not enabled by default.` 

As explicited in couchdb#1550 couchdb#1561 this is not the case, the session plugin is actually enabled by default.

## Possible Solutions
- Remove the note.
- Modify the note to highlight that the new plugin will break replication with endpoints using `require_valid_user` until a fix is found.

shimaore:
## Overview

The session authentication plugin is actually enabled in 2.2.0 and breaks existing replications with peers using `require_valid_user`.

## Testing recommendations

Documentations updates should be tested.

## GitHub issue number

Fixes #319

## Related Pull Requests

none

## Checklist

### DIFF
--- a/src/config/replicator.rst
+++ b/src/config/replicator.rst
@@ -229,4 +229,4 @@ Replicator Database Configuration
           auth_plugins = couch_replicator_auth_session,couch_replicator_auth_noop
 
         .. note::
-             In version 2.2, the session plugin is considered experimental and is not enabled by default.
+             In version 2.2.0, the session plugin is enabled by default and may break replication with peers using `require_valid_user`.


wohali:
@nickva Keep in mind that we are unable to update the documentation on the docs.couchdb.org website for 2.2.0 as that tag has landed.

As this will be fixed with the very next version of CouchDB that is released, there is no point in updating this documentation at all. We will mention the problem with 2.2.0 in the 2.2.1 (or whatever version) release notes.

@shimaore thanks again for the PR, sorry that we are unable to accept this.

PR #326, opened 2018-09-12
Split of Firewall and Erlang Configuration

ufobat:
While installing a CouchDB cluster I was struggeling over some points. With this PR I am hoping to emphasize what configurational steps are actually required required but where hidden in either the Firewall or Configuration Wizard section.

1. Within the part that describes the firewall requirements there was also quite a lot of information that describes *things* regarding Erlang. Therefore I created a subsection regardins all the Erlang *things*

2. Some parts are about testing the communication in/with Erlang, other parts are about actually configurating settings in`vm.args`. In the chapter about using the cluster wizard there was information about the ip|fqdn settings. I think that those settings are always required to be changed for all cluster installation.

3. The configuration curl calls for passwod and bind_address is using `couchdb@<this-nodes-ip-address>`. Previously IP or FQDN was allows in the vm.args file. I think that those 2 parts belong together so I made them consistent.

## Checklist


src/cluster/setup.rst
```diff
@@ -45,6 +45,28 @@ possible to force an Erlang application to use a specific port rage.
 This documentation will use the range TCP ``9100-9200``. Open up those ports in
 your firewalls and it is time to test it.
 
+Configure And Test The Communication With Erlang
```
flimzy: I suggest using normal "Title Case" capitalization here, to be consistent with the rest of the documentation.

> Configure and Test the Communication with Erlang

src/cluster/setup.rst
```diff
@@ -45,6 +45,28 @@ possible to force an Erlang application to use a specific port rage.
...
+================================================
+
+Make CouchDB use correct IP|FQDN and the open ports.
+----------------------------------------------------
+
+In file etc/vm.args change the the line ``-name couchdb@127.0.0.1`` to
```
flimzy: For consistency, "etc/vm.args" should be back-quoted:

> In file ``etc/vm.args`` change...

Also, you duplicated the word "the".

src/cluster/setup.rst
```diff
@@ -45,6 +45,28 @@ possible to force an Erlang application to use a specific port rage.
...
+``-name couchdb@<this-nodes-ip-address|FQDN>`` for each node which defines
+the node and must be seperate for each node. For clustered setup, each node in
```
flimzy: This sentence confuses me.  Do you mean "for each line which defines the node"?

And what does "must be separate for each node" mean?   A unique line per node?

ufobat:
any feedback regarding this change

src/cluster/setup.rst
```diff
@@ -45,6 +45,28 @@ possible to force an Erlang application to use a specific port rage.
 This documentation will use the range TCP ``9100-9200``. Open up those ports in
 your firewalls and it is time to test it.
 
+Configure and Test the Communication with Erlang
+================================================
+
+Make CouchDB use correct IP|FQDN and the open ports.
```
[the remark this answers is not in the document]
flimzy: Is an FQDN actually required? I would expect that the only requirement is a _resolvable hostname_, which should typically allow for relative hostnames in conjunction with the OS's search path, etc.

If my understanding is correct, can we replace FQDN with simply "host"?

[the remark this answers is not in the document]
wohali: Erlang bypasses `/etc/hosts` and local libresolv files, and interprets FQDNs different from "short names", so yes, it has to be an actual FQDN.

ufobat:
I've added the requested changes.

wohali:
@ufobat Thanks! We're waiting for GitHub to catch up with backlogged webhooks so that Travis can run and pass the checks - but this PR is good to go now.
 
Thank you for your contribution to CouchDB! :+1:

PR #335, opened 2018-10-03
Update responses and UI copy in Getting Started Guide

m90:
## Overview

I just gave the Getting Started Guide a go and noticed that some of the responses and UI elements that are being referred to are out of sync with a fresh 2.2.0 installation. This PR syncs the document with the values returned / used.

## Testing recommendations

Run a fresh install of CouchDB 2.2.0, then do what is described in the Getting Started Guide.

## Checklist

### DIFF
--- a/src/intro/tour.rst
+++ b/src/intro/tour.rst
@@ -41,10 +41,15 @@ The reply should look something like:
 
     {
         "couchdb": "Welcome",
+        "version": "2.2.0",
+        "git_sha":"2a16ec4",
+        "features": [
+            "pluggable-storage-engines",
+            "scheduler"
+        ],
         "vendor": {
             "name": "The Apache Software Foundation"
-        },
-        "version": "2.0.0"
+        }
     }
 
 Not all that spectacular. CouchDB is saying "hello" with the running version
@@ -58,7 +63,7 @@ All we added to the previous request is the _all_dbs string.
 
 The response should look like::
 
-    ["_replicator","_users"]
+    []
 
 Oh, that's right, we didn't create any databases yet! All we see is an empty
 list.
@@ -335,7 +340,7 @@ Edit the map function, on the right, so that it looks like the following:
 This is a JavaScript function that CouchDB runs for each of our documents as
 it computes the view. We'll leave the reduce function blank for the time being.
 
-Click "Run" and you should see result rows,
+Click "Save Document and then Build Index" and you should see result rows,
 with the various items sorted by price. This map function could be even more
 useful if it grouped the items by type so that all the prices for bananas were
 next to each other in the result set. CouchDB's key sorting system allows any


src/intro/tour.rst
```diff
@@ -58,7 +63,7 @@ All we added to the previous request is the _all_dbs string.
 
 The response should look like::
 
-    ["_replicator","_users"]
+    []
```
wohali: This means you didn't finish installation correctly. See http://docs.couchdb.org/en/stable/install/setup.html for the steps you missed.

Absolutely the response should be `["_replicator","_users"]`. If you want to put an admonishment in here with a link to the doc page I mentioned above, go for it. ([Use the RST syntax for linking](http://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#ref-role), not a link to the URL, please.)

m90: Ok, I see. What I did is just pull `couchdb:2.2.0` from docker hub, `run` it, and since I was able to read and write data, I assumed things were setup correctly, so adding a link to the setup page would definitely be a good idea I think.

That being said, after doing what I have been advised to in the setup doucment I will see the following:

```
➜  ~ curl http://127.0.0.1:5984/_all_dbs
["_global_changes","_replicator","_users"]
```

which still does not match the documented response. Should this be edited in order to contain the `_global_changes`?

wohali: Yes, sorry, `_global_changes` can be added to the response. Its creation is optional, but useful to highlight.

The Docker container doesn't do the initial creation of databases because that cannot be done until all nodes are joined to the cluster. If you're using Docker to create a 3-node cluster, you wouldn't want it to incorrectly create those databases locally on each node before you have a chance to merge the nodes together into a single cluster. It's a chicken-and-egg problem ;)

wohali: @m90 Would you have time to get this cleaned up soon? We are hoping to release 2.3.0 as soon as next week.

m90: @wohali Done so.

src/intro/tour.rst
```diff
@@ -58,7 +63,12 @@ All we added to the previous request is the _all_dbs string.
 
 The response should look like::
 
-    ["_replicator","_users"]
+    ["_global_changes","_replicator","_users"]
+
+.. note::
+    In case this returns an empty Array for you, it means you haven't finished
```
flimzy: Should "Array" be capitalized?  I would think no, it doesn't seem like a proper noun to me, but I may be wrong in this context?

wohali:
Hi @ufobat,

We can't merge until the build is fixed. You're currently getting this error:

```
Warning, treated as error:
/home/travis/build/apache/couchdb-documentation/src/intro/tour.rst:69:undefined label: install/setup (if the link has no caption the label must precede a section header)
```

The reference endpoint has changed to just `setup` now.

ufobat:
@m90, I think :-) not me.

PR #340, opened 2018-10-20
Add a paragraph about tuning couchjs

iblislin:
xref: apache/couchdb#1670

### DIFF
--- a/src/config/query-servers.rst
+++ b/src/config/query-servers.rst
@@ -59,6 +59,15 @@ Query Servers Definition
         javascript = /usr/bin/couchjs /usr/share/couchdb/server/main.js
         coffeescript = /usr/bin/couchjs /usr/share/couchdb/server/main-coffee.js
 
+    By default, ``couchjs`` limits the max runtime allocation to 64MiB.
+    If you run into out of memory issue in your ddoc functions,
+    you can adjust the memory limitation::
+
+        [query_servers]
+        javascript = /usr/bin/couchjs -S 536870912 /usr/share/couchdb/server/main.js ; 512 MiB
+
+    For more info about the available options, please consult ``couchjs -h``.
+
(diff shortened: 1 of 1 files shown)


src/config/query-servers.rst
```diff
@@ -59,6 +59,15 @@ Query Servers Definition
         javascript = /usr/bin/couchjs /usr/share/couchdb/server/main.js
         coffeescript = /usr/bin/couchjs /usr/share/couchdb/server/main-coffee.js
 
+    By default, ``couchjs`` limits the max runtime allocation in 64MiB.
```
flimzy: Should be "to 64MiB".

iblis17: fix

src/config/query-servers.rst
```diff
@@ -59,6 +59,15 @@ Query Servers Definition
         javascript = /usr/bin/couchjs /usr/share/couchdb/server/main.js
         coffeescript = /usr/bin/couchjs /usr/share/couchdb/server/main-coffee.js
 
+    By default, ``couchjs`` limits the max runtime allocation to 64MiB.
+    If you run into out of memory issue in your ddoc functions,
+    you can adjust the memory limitation::
+
+        [query_servers]
+        javascript = /usr/bin/couchjs -S 536870912 /usr/share/couchdb/server/main.js ; 512 MiB
```
wohali: I don't believe you can use `;` as a comment character in the ini file.

Can you move the comment into the previous paragraph instead, so people cutting and pasting this snippet won't break their installations? Thanks!

iblis17: I think `;` works correctly in my couchdb config file.
And there is another exaple: 
https://github.com/apache/couchdb/blob/master/test/random_port.ini

iblis17: https://github.com/apache/couchdb/blob/master/rel/overlay/etc/local.ini#L8

PR #354, opened 2018-11-16
Fix RST syntax errors

wohali:
Oops, some syntax errors in the Shard Maintenance section meant that any code blocks weren't being displayed.

### DIFF
--- a/src/api/database/find.rst
+++ b/src/api/database/find.rst
@@ -176,7 +176,7 @@ documents whose "director" field has the value "Lars von Trier".
 
     A simple selector, inspecting specific fields
 
-.. code:: json
+.. code-block:: javascript
 
     "selector": {
       "$title": "Live And Let Die"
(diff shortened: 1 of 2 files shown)


kxepal:
Hm..these two are actually the same. Or should be the same. What's the case?

garbados:
I'm getting the same error as Travis when I try `make html`, though `make check` throws an error that seems unrelated to this PR:

```
$ make check
python ext/linter.py src/
/home/garbados/code/couchdb-documentation/src/whatsnew/2.3.rst
  line 86 : trailing whitespace detected!
    If the legacy httpsd configuration is found in your ini file, this will 

Makefile:48: recipe for target 'check' failed
make: *** [check] Error 1
```

wohali:
@kxepal they're not, see: http://docs.couchdb.org/en/stable/cluster/sharding.html the code examples are not rendering at all

kxepal:
P.S. :wave: @wohali  (:

wohali:
@kxepal hi! nice to see you :) I saw you're busy with other projects now, was very surprised to see you pop up here :) you're welcome back anytime you want. We miss you!

PR #355, opened 2018-11-16
fix: remove trailing whitespace

garbados:
I noticed some errant whitespace was causing `make check` to fail while checking https://github.com/apache/couchdb-documentation/pull/354. This PR removes it.


wohali:
+1

PR #356, opened 2018-11-19
Improve CouchDB setup documentation, esp. for clustering

wohali:
## Overview

Our cluster setup instructions were a mess. Not only that, they were missing two very critical points: consistent UUID and consistent http secret across all nodes in a cluster. (Our setup wizard doesn't currently ensure these, so they must be done manually.)

If you are reviewing this PR, note that `src/setup/cluster.rst` was not just moved to a new directory, but completely overhauled.

I'd like to get this in before 2.3.0 releases.

## Testing recommendations

You could try setting up a new cluster using the information in `src/setup/cluster.rst` if you wanted.

## GitHub issue number

Addresses numerous requests that have come in through IRC/Slack, but no specific GH issue.

## Checklist


janl:
slight preference to have the wizard and api come before the manual setup, but great work overall. +1

wohali:
@janl It is, the setup wizard doesn't do enough, so manual steps are required before it can be used. The fully manual section is still at the bottom.

PR #358, opened 2018-11-22
Startup script uses COUCHDB_ARGS_FILE for vm.args path

natevw:
This seems to be a discrepancy from the start, see original pull requests:

* https://github.com/apache/couchdb/pull/1095 (code)
* https://github.com/apache/couchdb-documentation/pull/227 (docs)

~~Note also the script seems unhappy with relative paths, so I've also snuck a wording change alongside the main fix here.~~ <ins>I've filed https://github.com/apache/couchdb/issues/1757 instead, and kept this patch focused on the incorrect ENV variable name</ins>


wohali:
+1, thanks @natevw

PR #363, opened 2018-12-03
old favicon still turns up in local/offline docs

dch:
## Testing recommendations

I see this in 2 places - in `/docs/_static/` which this fixes, and in `./www/couchdb2/favicon.ico` just on FreeBSD port. I am not yet sure where this gets pulled in from.

## Checklist


wohali:
+1

PR #366, opened 2018-12-14
Add some query params to the GET /_all_dbs endpoint

nono:
## Overview

I've added documentation about some query parameters for the `GET /_all_dbs` endpoint. I've tested that these parameters work on CouchDB 2.3.0.

## Checklist


wohali:
Thanks for this!

PR #371, opened 2018-12-20
More how tos

wohali:
* Typos and fixes from @flimzy reviews of the last PR
* Migrated "Apache as a reverse proxy", called out HAProxy more prominently as preferred solution
* Adding pointer to more Erlang query server examples
* Add link to (very old) video on vhost/rewrite behaviour

There are only a few pages left; I will probably finish migrating the wiki tomorrow :dancing_women:

### DIFF
--- a/src/best-practices/reverse-proxies.rst
+++ b/src/best-practices/reverse-proxies.rst
@@ -322,7 +322,7 @@ Basic Configuration
 
 Here's a basic excerpt for using a ``VirtualHost`` block config to use Apache
 as a reverse proxy for CouchDB. You need at least to configure Apache with the
-``--enable-proxy --enable-proxy-http`` options and use a version equal or
+``--enable-proxy --enable-proxy-http`` options and use a version equal to or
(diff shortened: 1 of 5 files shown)


src/best-practices/reverse-proxies.rst
```diff
@@ -267,3 +306,45 @@ Certificates are issued by the LetsEncrypt certificate authority.
...
+    replication. The Apache CouchDB team strongly recommend the use of an
+    alternative reverse proxy such as ``haproxy`` or ``nginx``, as described
+    earlier in this section.
+
+Basic Configuration
+-------------------
+
+Here's a basic excerpt for using a ``VirtualHost`` block config to use Apache
+as a reverse proxy for CouchDB. You need at least to configure Apache with the
+``--enable-proxy --enable-proxy-http`` options and use a version equal or
```
flimzy:
```suggestion
``--enable-proxy --enable-proxy-http`` options and use a version equal to or
```

PR #378, opened 2019-01-18
Document 1x new and 2x undocumented shard endpoints

wohali:
New: apache/couchdb#1807 's `/{db}/_sync_shards`
Previously undocumented: `/{db}/_shards` and `/{db}/_shards/{docid}`. These have been around since 2.0.

### DIFF
--- a/src/api/database/shard.rst
+++ b/src/api/database/shard.rst
@@ -1,0 +1,223 @@
+.. Licensed under the Apache License, Version 2.0 (the "License"); you may not
+.. use this file except in compliance with the License. You may obtain a copy of
+.. the License at
+..
+..   http://www.apache.org/licenses/LICENSE-2.0
+..
+.. Unless required by applicable law or agreed to in writing, software
+.. distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+.. WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+.. License for the specific language governing permissions and limitations under
+.. the License.
+
+.. _api/db/shards:
+
+===============
+``/db/_shards``
+===============
+
+.. versionadded:: 2.0
+
+.. http:get:: /{db}/_shards
+    :synopsis: Displays the shard map layout of a database
+
+    The response will contain a list of database shards. Each shard will
+    have its internal database range, and the nodes on which replicas of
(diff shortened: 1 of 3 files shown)


src/api/database/shard.rst
```diff
@@ -0,0 +1,220 @@
...
+    :<header Accept: - :mimetype:`application/json`
+                     - :mimetype:`text/plain`
+    :>header Content-Type: - :mimetype:`application/json`
+                           - :mimetype:`text/plain; charset=utf-8`
+    :>json string range: The shard range in which the document is stored
+    :>json array nodes: List of nodes serving a replica of the shard
+    :code 200: Request completed successfully
+    :code 401: Read privilege required
+    :code 404: Database or document not found
+
```
flimzy: In general, I question the value of mentioning '500' errors in API documentation, since in theory, ti could happen anywhere, but since we do seem to mention that on other endpoints (including the one above), is it worth putting here?

src/api/database/shard.rst
```diff
@@ -0,0 +1,220 @@
...
+    :>header Content-Type: - :mimetype:`application/json`
+                           - :mimetype:`text/plain; charset=utf-8`
+    :>json boolean ok: Operation status. Available in case of success
+    :>json string error: Error type. Available if response code is ``4xx``
+    :>json string reason: Error description. Available if response code is
+      ``4xx``
+    :code 202: Request accepted
+    :code 400: Invalid database name
+    :code 401: CouchDB Server Administrator privileges required
+
```
flimzy: Same comment here re: 500 error.

flimzy: Testing indicates 404 is also possible here.

src/cluster/sharding.rst
```diff
@@ -121,6 +121,127 @@ guarantee `ACIDic consistency
...
+    {
+      "range": "60000000-7fffffff",
+      "nodes": [
+        "node1@127.0.0.1",
+        "node3@127.0.0.1",
+        "node4@127.0.0.1"
+      ]
+    }
+
+CouchDB shows us that the specific shard into which each of the two
```
flimzy:
```suggestion
CouchDB shows us the specific shard into which each of the two
```

PR #379, opened 2019-01-20
Improvements for bulk get documentation

flimzy:


src/api/database/bulk-api.rst
```diff
@@ -473,11 +473,15 @@ Sending multiple queries to a database
...
     :query boolean revs: Give the revisions history
-    :<json array docs: List of document objects, with ``id``, and optionnaly
+    :<json array docs: List of document objects, with ``id``, and optionally
       ``rev`` and ``atts_since``
     :>header Content-Type: - :mimetype:`application/json`
-    :>json object results: the documents, with the additionnal ``_revisions``
-      property that lists the parent revisions if ``revs=true``
+    :>json object results: an array of results for each requested document/rev
+    pair. ``id`` key lists the requested document ID, ``docs`` contains a
+    single-item array of objects, each of which has either an ``error`` key and
```
flimzy: Can anyone confirm that this is indeed always a single-item array? I've yet to find any query that produces more than a single item in this array.

wohali: FYI This whole section is missing the multipart content-type acceptance.

https://github.com/apache/couchdb/blob/16e6af4cf8ec268f7f7eef0f78b3cf0ec3b6f3bd/src/chttpd/src/chttpd_db.erl#L547-L549

Try accepting content-type `multipart/mixed` on a doc with attachments.

https://github.com/apache/couchdb-chttpd/pull/33 has the full history of the PR for this functionality, and it indeed does look like, in the JSON case, you get a single item array all the time, even for docs with multiple revs. But the code has an iterative loop in it, which possibly COULD include multiple docs - I just don't see how.

More history here: https://issues.apache.org/jira/browse/COUCHDB-2310

src/api/database/bulk-api.rst
```diff
@@ -473,11 +473,15 @@ Sending multiple queries to a database
...
     :<header Content-Type: :mimetype:`application/json`
     :query boolean revs: Give the revisions history
-    :<json array docs: List of document objects, with ``id``, and optionnaly
+    :<json array docs: List of document objects, with ``id``, and optionally
       ``rev`` and ``atts_since``
     :>header Content-Type: - :mimetype:`application/json`
-    :>json object results: the documents, with the additionnal ``_revisions``
-      property that lists the parent revisions if ``revs=true``
+    :>json object results: an array of results for each requested document/rev
+    pair. ``id`` key lists the requested document ID, ``docs`` contains a
```
wohali: this line (and the ones following it) are improperly indented, breaking the build. FYI.

PR #384, opened 2019-01-24
Config cluster ref fix

### LINKED ISSUE #383: Broken cross-reference in [cluster] configuration
In the `q` part of the "Configuring Clustering" documentation page (<https://docs.couchdb.org/en/2.3.0/config/cluster.html#cluster/q>) there is a note "The value of q can also be overridden on a per-DB basis, at DB creation time." followed by a "See also" callout.

The contents of that call-out are broken, and do not link to anything. Instead the box literally says:

> httpdomain:put:PUT /{db} </{db}>

(Compare to the `placement` section where a nice link to "Placing a database on specific nodes" is provided in a similar callout.)

jaydoane:
## Overview

This fixes the broken reference to the PUT db API

## Testing recommendations

`make html` and open `build/html/config/cluster.html` locally to confirm link works

Not sure if this is really passing though?
```
$ make check
python3 ext/linter.py src/
$ echo $?
0
```

## GitHub issue number

FIxes #383 

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 ext/*.pyc
 ext/__pycache__/
 build/
+venv
 *~
 *.swp
 *.bak
--- a/README.md
+++ b/README.md
@@ -10,7 +10,9 @@ You can view the latest rendered build of this content at:
 Install Python 2.7+ and pip. Then:
 
 ```sh
-$ sudo pip install Sphinx sphinx_rtd_theme
+$ virtualenv venv
+$ . ./venv/bin/activate
+$ pip install -r requirements.txt
 $ make html # builds the docs
 $ make check # syntax checks the docs
 ```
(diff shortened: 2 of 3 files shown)


jaydoane:
@wohali I tried to create a link directly to the [PUT anchor](http://docs.couchdb.org/en/latest/api/database/common.html#put--db) but couldn't figure it out. Is there are way to do so?

wohali:
@jaydoane We use an old version of https://sphinxcontrib-httpdomain.readthedocs.io/en/stable/ for the http stuff. (At some point we should unbundle our version under `ext/` and just use `pip` to install it.)

It's implemented as a Sphinx domain, so http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html should help as well.

```rst
:http:put:`PUT /{db} </{db}>`
```

should work.

jaydoane:
@wohali yes, that was the magic -- thanks!

PR #392, opened 2019-02-07
feat: 2.3.1 release notes

janl:


src/whatsnew/2.3.rst
```diff
@@ -107,6 +107,43 @@ Upgrade Notes
...
+=============
+
+Features
+--------
+
+.. rst-class:: open
+
+* :ghissue:`1811`: Add new ``/{db}/_sync_shards`` endpoint (admin-only).
+* :ghissue:`1870`: Update to mochiweb 2.19.0. See also :ghissue:`1875`.
+* :ghissue:`1875`: Refuse building with known bad versions of Erlang.
```
flimzy: And would "unsupported versions of Erlang" make more sense than "known bad"?

src/whatsnew/2.3.rst
```diff
@@ -107,6 +107,43 @@ Upgrade Notes
 * :ghissue:`1680`: The embedded version of ``rebar`` used to build CouchDB has been
   updated to the last version of ``rebar2`` available. This assists in building on
   non-x86 platforms.
+* :ghissue:`1875`: Refuse building with known bad versions of Erlang.
```
flimzy: Is this supposed to be here? It's also duplicated below.

src/whatsnew/2.3.rst
janl: good question, unsupported doesn’t have the same weight as known bad. The former implies we might be just too lazy to support them, while the latter says “while you might be able to run this, you will run into issues sooner or later so better don’t”

src/whatsnew/2.3.rst
janl: Yeah, I want it both in the upgrade nodes and the per-version section.

src/whatsnew/2.3.rst
flimzy: It's a minor issue. But "known bad" sounds (to me) like there are known bugs in Erlang, not simply an incompatibility.  Maybe "incompatible versions of Erlang"?  Or "known unsupported"?  Or stick with "known bad". :)  The risk of confusing people with a changelog is relatively low, as the people who read these are usually more technical than the average user.

janl: > But "known bad" sounds (to me) like there are known bugs in Erlang,

That is exactly the situation, CouchDB will work and run with these versions, but bugs in erlang can cause issues when using CouchDB.

janl: > It's a minor issue.

CouchDB has always been bout sweating the details, so I very much appreciate this ;)

PR #395, opened 2019-03-05
Include note about enabling couch_peruser.

foxmulder900:
## Overview
There is no indication in the documentation that the existence of the `_users` database is a prerequisite to enabling `couch_peruser`. This note would have saved me quite a bit of time. 

Source of note is here: 
https://github.com/apache/couchdb/issues/749#issuecomment-322057875

## Checklist

### DIFF
--- a/src/config/couch-peruser.rst
+++ b/src/config/couch-peruser.rst
@@ -33,6 +33,9 @@ couch_peruser Options
 
         [couch_peruser]
         enable = false
+        
+    .. note::
+        The ``_users`` database must exist before couch_peruser can be enabled.
 
     .. config:option:: delete_dbs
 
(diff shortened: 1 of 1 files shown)


wohali:
HI @foxmulder900 , thanks for the note.

Actually, creating the `_users` database **is** in the instructions for setting up CouchDB, which I guess you skipped:

* http://docs.couchdb.org/en/latest/setup/single-node.html
* http://docs.couchdb.org/en/latest/setup/cluster.html

I'll accept this PR, but it's a bit superfluous, in that people should be reading and following the set up instructions.

foxmulder900:
Thanks!

I would like to note that the documentation you linked describes setting up `_users` at runtime. Which is exactly what I did, but is too late if `couch_peruser` is enabled via `local.ini`. 

My experience was that the `_users` database had to be created as described, *then* an additional call to enable couch_peruser had to be made. Otherwise the setting was left disabled.

PR #397, opened 2019-03-05
Add latest RFC draft for FDB revision storage

kocolosk:
## Overview

RFC for the storage of revision metadata in FoundationDB.

## Testing recommendations

N/A

## GitHub issue number

N/A

## Related Pull Requests

N/A

## Checklist

### DIFF
--- a/rfcs/001-fdb-revision-metadata-model.md
+++ b/rfcs/001-fdb-revision-metadata-model.md
@@ -1,0 +1,179 @@
+[NOTE]: # ( ^^ Provide a general summary of the RFC in the title above. ^^ )
+
+# Introduction
+
+This is a proposal for the storage of document revision history metadata as a
+set of KVs in FoundationDB.
+
+## Abstract
+
+This design stores each edit branch as its own KV, and all of the edit branches
+are stored separately from the actual document data. Document reads can avoid
+retrieving this information, while writes can avoid retrieving the document
+body.
+
+## Requirements Language
+
+[NOTE]: # ( Do not alter the section below. Follow its instructions. )
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC
+2119](https://www.rfc-editor.org/rfc/rfc2119.txt).
+
+## Terminology
+
+[TIP]:  # ( Provide a list of any unique terms or acronyms, and their
+definitions here.)
+
+---
+
+# Detailed Description
+
+The size limits in FoundationDB preclude storing the entire revision tree as a
+single value; in pathological situations the tree could exceed 100KB. Rather, we
+propose to store each edit *branch* as a separate KV. Specifically, we create a
+"revisions" subdirectory in each database directory to store the revision trees
+with keys and values that look like
+
+`(“revisions”, DocID, NotDeleted, RevFormat, RevPosition, RevHash) =
+(Versionstamp, [ParentRev, GrandparentRev, …])`
+
+where the individual elements of the key and value are defined as follows:
+- `DocID`: the document ID
+- `NotDeleted`: `\x00` if the leaf of the edit branch is deleted, `\x01`
+  otherwise
+- `RevFormat`: enum for the revision encoding being used, start at `\x01` with
+  this proposal
+- `RevPosition`: positive integer encoded using standard tuple layer encoding
+  (signed, variable-length, order-preserving)
+- `RevHash`: 16 bytes uniquely identifying this revision
+- `Versionstamp`: the FoundationDB
+  [versionstamp](https://apple.github.io/foundationdb/api-python.html#fdb.tuple.Versionstamp)
+  associated with the last transaction that modified the document (NB: not
+  necessarily the last edit to *this* branch).
+- `[ParentRev, GrandparentRev, ...]`: 16 byte identifiers of ancestors, up to
+  1000 by default
+
+## Limits
+
+In order to stay compatible with FoundationDB size limits we need to prevent
+administrators from increasing `_revs_limit` beyond what we can fit into a
+single value. Suggest **4000** as a max.
+
+## Update Path
+
+Multiple edit branches on a document are largely independent of one another in
+this design, but some coordination is required around the `Versionstamp`. Recall
+that the `_changes` feed includes each document exactly once, so we do not want
+to be able to extend different edit branches in parallel and end up adding both
+stamps to the feed. We address this by storing the `Versionstamp` only on the
+so-called "winning" branch. Other branches set this to null.
+
+If a writer comes in and tries to extend a losing edit branch, it will find the
+`Versionstamp` to be null and will do an additional edit branch read to retrieve
+the winning branch. It can then compare both branches to see which one will be
+the winner following that edit, and can assign the new `Versionstamp` to that
+branch accordingly.
+
+A writer attempting to delete the winning branch (i.e., setting `NotDeleted` to
+0) will need to read two contiguous KVs, the one for the winner and the one
+right before it. If the branch before it will be the winner following the
+deletion then we move the storage of the new `Versionstamp` to it accordingly.
+If the tombstoned branch remains the winner for this document then we only
+update that branch.
+
+A writer extending the winning branch with an updated document (the common case)
+will proceed reading just the one branch.
+
+Summarizing the performance profile:
+- Extending a losing branch: 2 KVs, 2 roundtrips
+- Deleting the winning branch: 2 KVs, 1 roundtrip
+- Extending the winning branch: 1 KV, 1 roundtrip
+- `new_edits=false` update: `<N>` KVs, 1 roundtrip
+
+# Advantages
+
+We can read a document revision without retrieving the revision tree, which in
+the case of frequently-edited documents may be larger than the doc itself.
+
+We ensure that an interactive document update against the winning branch only
+needs to read the edit branch KV against which the update is being applied, and
+it can read that branch immediately knowing only the content of the edit that is
+being attempted (i.e., it does not need to read the current version of the
+document itself). The less common scenario of updating a losing branch is only
+slightly less efficient, requiring two roundtrips.
+
+Interactively updating a document with a large number of edit branches is
+therefore dramatically cheaper, as no more than two edit branches are read or
+modified regardless of the number of branches that exist, and no tree merge
+logic is required.
+
+Including `NotDeleted` in the key ensures that we can efficiently accept the
+case where we upload a new document with the same ID where all previous edit
+branches have been deleted; i.e. we can construct a key selector which
+automatically tells us there are no `deleted=false` edit branches.
+
+The `RevFormat` enum gives us the ability to evolve revision history storage
+over time, and to support alternative conflict resolution policies like Last
+Writer Wins.
+
+Access to `Versionstamp` ensures we can clear the old entry in the `by_seq`
+space during an edit. The `set_versionstamped_value` API is used to store this
+value automatically.
+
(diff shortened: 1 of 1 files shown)


rfcs/001-fdb-revision-metadata-model.md
```diff
@@ -0,0 +1,179 @@
...
+  associated with the last transaction that modified the document (NB: not
+  necessarily the last edit to *this* branch).
+- `[ParentRev, GrandparentRev, ...]`: 16 byte identifiers of ancestors, up to
+  1000 by default
+
+## Limits
+
+In order to stay compatible with FoundationDB size limits we need to prevent
+administrators from increasing `_revs_limit` beyond what we can fit into a
+single value. Suggest **4000** as a max.
```
rnewson: I'd also be fine with not allowing higher than the longstanding default of 1000, or even of allowing this to alter at all.

kocolosk: I think there's a clear case to be made for allowing users to lower it if they're not interested in replication and they update documents frequently.

I'm OK to limit to 1000 if that's the consensus.

kocolosk:
I thought of two additional bits while working through the `_changes` feed design:

1. Following Apple's Record Layer paper it's wise to have an "incarnation" number prefixing each `Versionstamp` to account for the possibility of a single database being relocated to a different FoundationDB cluster (with versionstamps that are uncorrelated with the previous cluster).
1. The `_changes` feed is often requested with `?style=all_docs`; it would be useful to record the count of edit branches in the `_changes` index so we could avoid an unnecessary read of the "revisions" subspace here in the common case of a single edit branch. We can rely on the fact that the number of edit branches only changes during a `new_edits=false` update, at which time the client has the full list of branches, to store the count of branches in the value of the winning edit branch.

I'll look to update the RFC with these concepts shortly.

kocolosk:
I've updated the document to cover both the `Incarnation` addition and the `BranchCount` addition.

One thing I'm not sure about is the placement of `RevFormat`. Does it really need to be in the key, or is it better suited as the header of the value? I had it in the key thinking that e.g. a Last Writer Wins resolution policy wouldn't even have `RevPosition` or `RevHash`, but I also think there ought to be some header to aid in decoding the value format, and I'm not sure if I'm maybe overloading `RevFormat` too much in relying on it to perform that function.

rfcs/001-fdb-revision-metadata-model.md
rnewson: Now that internal replication will not be done the same way there is less risk to lowering revs limit (at least, one fewer side-effect where eventual consistency of the 3 replicas themselves fails). Hopefully others will chime in with their views.

janl: It’s rather edge-casey, but allowing to lower revs_limit has proved beneficial in a handful of cases in the past. I would be okay with a fixed lower bound that is > 1 if we fix it at, say 10, which would still solve the edge-case, but don’t suggest the flawed understanding of revs_limit eliminating tombstones.

That said, it might be easier to keep the default and keep it configurable for the time being. I’d argue against hardcoding this.

rfcs/001-fdb-revision-metadata-model.md
```diff
@@ -0,0 +1,200 @@
...
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC
+2119](https://www.rfc-editor.org/rfc/rfc2119.txt).
+
+## Terminology
+
+[TIP]:  # ( Provide a list of any unique terms or acronyms, and their
+definitions here.)
```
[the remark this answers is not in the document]
kocolosk: Great idea, was thinking the same this morning regarding the sequence as a combination of incarnation and versionstamp. I'll start there and we can build up as we go along.

kocolosk:
I think this RFC would benefit from some additional discussion of the case when writers are racing to create the first edit of a document. I believe everything we're doing is still correct and needs no modification, but this is subtle enough that it should be called out explicitly.

The issue is that we're relying on the fact that every modification to a document reads the KV for the winning edit branch and modifies it, so FoundationDB will correctly identify concurrent transactions updating the same document as conflicting. But when we're making the first edit on a document there is no winning edit branch KV to read.

If an edit comes in without specifying any parent revision our layer code will still try to check for that winning edit branch KV (and fail the edit with a conflict if one exists). It'll need to do the check as a `get_range_startswith` operation with `reverse=true` and `limit=1`. If it finds a null result set it will assume that the insert is OK. If someone else wins the race and inserts the document then this should still trigger FoundationDB's conflict checking algorithm.

kocolosk:
Another topic left unhandled here is whether to continue with the practice of extending the winning tombstone edit branch when inserting a document where all edit branches were previously deleted. It is certainly possible to do so, and fits naturally into the documented edit path. But the user-facing implications of that practice are fairly subtle as it seems like this deserves its own discussion.

I'm also anticipating that we're going to want the edit branch metadata to record the size of the document body at that leaf so we can update the database size with each transaction. I suppose I could add that now, otherwise we'll have a dedicated discussion on that metadata soon and we can update the RFC then.

rfcs/001-fdb-revision-metadata-model.md
```diff
@@ -0,0 +1,215 @@
...
+- `DocID`: the document ID
+- `NotDeleted`: `\x26` if the leaf of the edit branch is deleted, `\x27`
+  otherwise (following tuple encoding for booleans)
+- `RevPosition`: positive integer encoded using standard tuple layer encoding
+  (signed, variable-length, order-preserving)
+- `RevHash`: 16 bytes uniquely identifying this revision
+- `RevFormat`: enum for the revision encoding being used to enable schema
+  evolution
+- `Sequence`: the sequence of the last transaction that modified the document
+  (NB: not necessarily the last edit to *this* branch).
```
garrensmith: Would you mind just elaborating on what you mean by `NB: not necessarily the last edit to *this* branch`?

kocolosk: This is covered in the "Update Path" section, but I can summarize here. An update that extends a losing edit branch will still modify the winning edit branch KV as well, setting the Sequence (which is only stored on the winning branch) to the sequence of that transaction.

garrensmith: Ok great. Thanks for clarifying.

davisp:
Just to make sure we're both on the same page I went through and wrote up a rough outline to the old and new update algorithms. One thing we'll probably want to call out is that weird edge cases may not end up exactly the same (i.e., an update that is a conflict and has missing stubs might return a different error than previously).

https://gist.github.com/davisp/7ab359038b6552296dac515a20f47087

The big issue I see though is that replicated updates require reading all revisions with this model. Given that the RFC calls out storing revision branches separately since we can exceed FDB transaction limits this seems like a bit of an issue. In a pathological case we could get to a position where we can update a document interactively but not via replication.

The first obvious approach here is to start failing writes once we get above some threshold of revision counts. However, this could get weird with having a single poison doc preventing replication into the database.

The second vague not-at-all thought out would be to switch the NotDeleted field to a RevStatus that is an enum of (`?REV_MISSING`, `?DELETED_TRUE`, `?DELETED_FALSE`) with that sort order and then the value is just a previous revision instead of the whole path. This change then includes an extra write for interactive updates when we add a new key for the `?REV_MISSING` status of the branch that was extended. For the replicated changes we'd also extend our read round trips to some complicated average of updates between replications with the bound being `revs_limit` read round trips which I'm not super fond of.

I'm also not a huge fan of the complexity around maintaining the adjacency list consistency guarantees and the subtle issues on that. Its been hard enough as a complete tree in RAM that the subtlety around adjacency list updates is a bit worrisome.

kocolosk:
> Given that the RFC calls out storing revision branches separately since we can exceed FDB transaction limits this seems like a bit of an issue

I was worried more about the 100K value size limit than the 10M transaction limit.

But OK, we could theoretically have a pathological case where the rev tree grows that large. Two points:

1) It turns out that FDB doesn’t actually care about the size of the values returned; all it cares about for the 10M limit is the size of the transaction being committed, which I certainly hope is smaller :)

2) If push comes to shove, we can always split a replicated edit into multiple transactions without breaking any contract in our API.

davisp:
@kocolosk I've put together a simple prototype covering the interactive side of this work here:

https://github.com/apache/couchdb/tree/prototype/rfc-001-revision-metadata-model

Notable things that are missing are:

  1. Replicated changes updates
  2. Doc open options like `?conflicts=true`, etc
  3. Tests beyond the most basic of basic "Yep, can write a doc"

I kept a few notes here:

https://github.com/apache/couchdb/blob/38c0f69410a31507309aef5a86a249efbca6ba7a/RFC001_NOTES.md

I think the biggest thing that I found missing is that we'll want to track the size of doc bodies in the revision info structure so we can properly account for that.

Also, the logic around Mutating the winning revisions and so on sure does get super subtle and involved lots of "Think Really Hard" steps that I'm not super fond of. Although, it suddenly occurs to me out of nowhere that I can probably go back and simplify that considerably by reusing couch_key_tree and friends. I'll take a crack at that approach and will see how well that works. That could be nice as then we could reuse a lot of the fdb logic for both replicated and interactive updates.

kocolosk:
> I think the biggest thing that I found missing is that we'll want to track the size of doc bodies in the revision info structure so we can properly account for that.

Ah! I'd made a mental note to add that several weeks ago and just forgot about it. I agree it's going to be needed.

davisp:
Another API edge case to consider is that we can currently extend a revision branch an arbitrary number of revisions even when not doing `new_edits=false`. I suddenly remembered this whacky corner case and realized I currently have a "bug" due to a subtle assumption that the provided revision in a `new_edits=true` update only ever specifies a single revision.

```
$ curl -X PUT http://127.0.0.1:5984/foo -d '{}'
{"ok":true,"id":"foo","rev":"1-967a00dff5e02add41819138abb3284d"}

$ cat foo.json
{
    "_id": "foo",
    "_revisions": {
        "start": 4,
        "ids": [
            "967a00dff5e02add41819138abb3284a",
            "967a00dff5e02add41819138abb3284b",
            "967a00dff5e02add41819138abb3284c",
            "967a00dff5e02add41819138abb3284d"
        ]
    },
    "blargh": "hgralb"
}

$ curl -X PUT http://127.0.0.1:5984/foo -d @foo.json
{"ok":true,"id":"foo","rev":"5-4c1a3fa652c789437f9d509f69521fdb"}
```

davisp:
Also second fun fact is that only works when there aren't any attachments or validate_doc_update functions set. CouchDB actually has the same "bug" when we go to load the previous doc from disk in prep_and_validate_updates.

Also it occurs to me that we can't just go to the end of the revision list. We'd have to try every revision provided to see if any of them match an existing revision that could be extended. Which... seems like not something we should add at this point?

kocolosk:
Whoa -- that's pretty messed up. I don't think we should carry it forward.

rfcs/001-fdb-revision-metadata-model.md
```diff
@@ -0,0 +1,215 @@
...
+[TIP]:  # ( Provide a list of any unique terms or acronyms, and their
+definitions here.)
+
+`Versionstamp`: a 12 byte, unique, monotonically (but not sequentially)
+increasing value for each committed transaction. The first 8 bytes are the
+committed version of the database. The next 2 bytes are monotonic in the
+serialization order for transactions. The final 2 bytes are user-defined and can
+be used to create multiple versionstamps in a single transaction.
+
+`Incarnation`: a single byte, monotonically increasing value specified for each
```
garrensmith: @ryanworl brings up a good point on the [Mango RFC](https://github.com/apache/couchdb-documentation/pull/407#discussion_r279957158) that the incarnation should rather be an integer.

PR #402, opened 2019-03-29
Update security.rst

tabeth:
Someone following the example here will mistakenly think `NODENAME` has been set. From a beginner's perspective I think `_local` is an OK default.

## Overview

From a beginner's perspective having `NODENAME` set will allow them to follow the example more closely. 

## Checklist

### DIFF
--- a/src/intro/security.rst
+++ b/src/intro/security.rst
@@ -84,6 +84,7 @@ Let's do another walk through the API using `curl` to see how CouchDB behaves
 when you add admin users. ::
 
     > HOST="http://127.0.0.1:5984"
+    > NODENAME="_local"
     > curl -X PUT $HOST/database
     {"ok":true}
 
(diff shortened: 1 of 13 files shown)


wohali:
Hm, ok, except that this means you can't be talking through a load balancer, and you will only ever be able to query the info of the node you're speaking to.

If you want to improve this further, do you want to put in a short description of what `NODENAME` is, where to find it, and that we also support the `_local` alias to mean the specific node you're talking to (pointing out the load balancer problem)?

tabeth:
Good point. I'll incorporate those things as well as link to the relevant items.

tabeth:
Wow, @wohali, apologies. I thought I made these changes months ago! In any case, I've added a link to the Node Management page, as well as a quick explanation to what `_local`  and `NODENAME` do in this context.

src/intro/security.rst
```diff
@@ -99,6 +100,13 @@ As per the :ref:`_config <api/config>` API's behavior, we're getting
 the previous value for the config item we just wrote. Since our admin user
 didn't exist, we get an empty string.
 
+Please note that ``_local`` serves as an  alias for the local node name, so 
+for all configuration URLs, ``NODENAME`` may be set to ``_local``, to interact 
+with the local node’s configuration.
+
+.. seealso::
+    :ref:`Node Management <stable/cluster/node>`
```
wohali: Unfortunately, this isn't a valid reference. Can you please fix this?

wohali:
Hey @tabeth you need to remove the `stable/`, that's just an artefact of how ReadTheDocs publishes our content on docs.couchdb.org for us. (It's a version label - it could be `latest/` or `2.3.1/` or so on.)

Just use `<cluster/nodes>`.

tabeth:
Should be good to go!

wohali:
@tabeth Congratulations on your first contribution to Apache CouchDB! :confetti_ball: :frog: :gem:

tabeth:
@wohali Expect more where that came from!

PR #403, opened 2019-04-01
RFC for document storage

kocolosk:
## Overview

This document describes a data model for storing JSON documents as key-value
pairs in FoundationDB. It includes a discussion of storing multiple versions of
the document, each identified by unique revision identifiers, and discusses some
of the operations needed to query and modify these documents.

### DIFF
--- a/rfcs/004-document-storage.md
+++ b/rfcs/004-document-storage.md
@@ -85,7 +85,7 @@ identifying the path to a final leaf value. For example, the object
 would be represented by a key-value pair of
 
 ```
-pack({"foo", "bar", "baz"}) = pack(123)
+pack({"foo", "bar", "baz"}) = pack({123})
 ```
 
 An array of N elements is represented by N distinct key-value pairs, where the
--- a/rfcs/004-document-storage.md
+++ b/rfcs/004-document-storage.md
@@ -106,11 +106,11 @@ position of the value within the array. As an example:
 becomes
 
 ```
-pack({"states", 0}) = "MA"
-pack({"states", 1}) = "OH"
-pack({"states", 2}) = "TX"
-pack({"states", 3}) = "NM"
-pack({"states", 4}) = "PA"
+pack({"states", 0}) = pack({"MA"})
+pack({"states", 1}) = pack({"OH"})
+pack({"states", 2}) = pack({"TX"})
+pack({"states", 3}) = pack({"NM"})
+pack({"states", 4}) = pack({"PA"})
 ```
 
 More details on the encodings in the FoundationDB Tuple Layer can be found in
--- a/rfcs/004-document-storage.md
+++ b/rfcs/004-document-storage.md
@@ -88,6 +88,11 @@ would be represented by a key-value pair of
 pack({"foo", "bar", "baz"}) = pack({123})
 ```
 
+Clients SHOULD NOT submit objects containing duplicate keys, as CouchDB will
+only preserve  the last occurence of the key and will silently drop the other
+occurrences. Similarly, clients MUST NOT rely on the ordering of keys within an
+Object as this ordering will generally not be preserved by the database.
+
 An array of N elements is represented by N distinct key-value pairs, where the
 last element of the tuple key is an integer representing the zero-indexed
 position of the value within the array. As an example:
--- a/rfcs/004-document-storage.md
+++ b/rfcs/004-document-storage.md
@@ -1,0 +1,244 @@
+---
+name: Formal RFC
+about: Submit a formal Request For Comments for consideration by the team.
+title: 'JSON document storage in FoundationDB'
+labels: rfc, discussion
+assignees: ''
+
+---
+
+[NOTE]: # ( ^^ Provide a general summary of the RFC in the title above. ^^ )
+
+# Introduction
+
+This document describes a data model for storing JSON documents as key-value
+pairs in FoundationDB. It includes a discussion of storing multiple versions of
+the document, each identified by unique revision identifiers, and discusses some
+of the operations needed to query and modify these documents.
+
+## Abstract
+
+The data model maps each "leaf" JSON value (number, string, true, false, and
+null) to a single KV in FoundationDB. Nested relationships are modeled using a
+tuple structure in the keys. Different versions of a document are stored
+completely independently from one another. Values are encoded using
+FoundationDB's tuple encoding.
+
+The use of a single KV pair for each leaf value implies a new 100KB limit on
+those values stored in CouchDB documents. An alternative design could split
+these large (string) values across multiple KV pairs.
+
+Extremely deeply-nested data structures and the use of long names in the nesting
+objects could cause a path to a leaf value to exceed FoundationDB's 10KB limit
+on key sizes. String interning could reduce the likelihood of this occurring but
+not eliminate it entirely. Interning could also provide some significant space
+savings in the current FoundationDB storage engine, although the introduction of
+key prefix elision in the Redwood engine should also help on that front.
+
+FoundationDB imposes a hard 10MB limit on transactions. In order to reserve
+space for additional metadata, user-defined indexes, and generally drive users
+towards best practices in data modeling this RFC proposes a **1 MiB** limit on
+document sizes going forward.
+
+## Requirements Language
+
+[NOTE]: # ( Do not alter the section below. Follow its instructions. )
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.txt).
+
+## Terminology
+
+---
+
+# Detailed Description
+
+## Value Encoding
+
+The `true` (`\x27`), `false` (`\x26`) and `null` (`\x00`) values each have a
+single-byte encoding in FoundationDB's tuple layer. Integers are represented
+with arbitrary precision (technically, up to 255 bytes can be used).
+Floating-point numbers use an IEEE binary representation up to double precision.
+More details on these specific byte codes are available in the [FoundationDB
+documentation](https://github.com/apple/foundationdb/blob/6.0.18/design/tuple.md).
+
+Unicode strings must be encoded into UTF-8. They are prefixed with a `\x02`
+bytecode and are null-terminated. Any nulls within the string must be replaced
+by `\x00\xff`. Raw byte strings have their own `\x01` prefix and must follow the
+same rules regarding null bytes in the string. Both are limited to 100KB.
+
+An object is decomposed into multiple key-value pairs, where each key is a tuple
+identifying the path to a final leaf value. For example, the object
+
+```
+{
+    "foo": {
+        "bar": {
+            "baz": 123
+        }
+    }
+}
+```
+
+would be represented by a key-value pair of
+
+```
+pack({"foo", "bar", "baz"}) = pack(123)
+```
+
+An array of N elements is represented by N distinct key-value pairs, where the
+last element of the tuple key is an integer representing the zero-indexed
+position of the value within the array. As an example:
+
+```
+{
+    "states": ["MA", "OH", "TX", "NM", "PA"]
+}
+```
+
+becomes
+
+```
+pack({"states", 0}) = "MA"
+pack({"states", 1}) = "OH"
+pack({"states", 2}) = "TX"
+pack({"states", 3}) = "NM"
+pack({"states", 4}) = "PA"
+```
+
+More details on the encodings in the FoundationDB Tuple Layer can be found in
+the [design
+documentation](https://github.com/apple/foundationdb/blob/6.0.18/design/tuple.md).
+
+## Document Subspace and Versioning
+
+Document bodies will be stored in their own portion of the keyspace with a fixed
+single-byte prefix identifying the "subspace". Each revision of a document will
+be stored separately without term sharing, and the document ID and revision ID
+are baked into the key. The structure looks like this
+
+```
+{DbName, ?DOCUMENTS, DocID, NotDeleted, RevPos, RevHash} = RevisionMetadata
+{DbName, ?DOCUMENTS, DocID, NotDeleted, RevPos, RevHash, "foo"} = (value for doc.foo)
+et cetera
+```
+
+where `RevisionMetadata` includes at the minium an enum to enable schema
+evolution for subsequent changes to the document encoding structure, and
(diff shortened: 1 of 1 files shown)


rfcs/004-document-storage.md
```diff
@@ -0,0 +1,244 @@
...
+- it lends itself nicely to sub-document operations, e.g. apache/couchdb#1559
+- it optimizes the creation of Mango indexes on existing databases since we only
+  need to retrieve the value(s) we want to index
+- it optimizes Mango queries that use field selectors
+
+The disadvantages of this approach are that it uses a larger number of key-value
+pairs and has a higher overall storage overhead from the repeated common key
+prefixes. The new FoundationDB storage engine should eliminate some of the
+storage overhead.
+
```
janl:
```suggestion
As per the FoundationDB discussion about being able to co-locate compute operations with data storage servers/nodes](https://forums.foundationdb.org/t/feature-request-predicate-pushdown/954/6), if we were to make use of this hypothetical feature, we’d not get a guarantee of entire documents being co-located on one storage node, requiring us to do extra work should we want to, say, assemble a full `doc` to send to a map function. JS views would have a harder time, while Mango indexes with their explicit field declarations might get around this particular complexity more easily. For now, this is recorded here so we don’t forget track of this later.


```

rfcs/004-document-storage.md
```diff
@@ -0,0 +1,246 @@
...
+            "baz": 123
+        }
+    }
+}
+```
+
+would be represented by a key-value pair of
+
+```
+pack({"foo", "bar", "baz"}) = pack(123)
```
rnewson: at least in @davisp's work on erlfdb so far, packing only happens for tuples. so should this be `pack({123})` or should erlfdb allow packing of primitives? I think Paul is reflecting the API's elsewhere so our choice may be forced here as we'd like our data to be readable with those API's.

rfcs/004-document-storage.md
```diff
@@ -0,0 +1,246 @@
...
+Floating-point numbers use an IEEE binary representation up to double precision.
+More details on these specific byte codes are available in the [FoundationDB
+documentation](https://github.com/apple/foundationdb/blob/6.0.18/design/tuple.md).
+
+Unicode strings must be encoded into UTF-8. They are prefixed with a `\x02`
+bytecode and are null-terminated. Any nulls within the string must be replaced
+by `\x00\xff`. Raw byte strings have their own `\x01` prefix and must follow the
+same rules regarding null bytes in the string. Both are limited to 100KB.
+
+An object is decomposed into multiple key-value pairs, where each key is a tuple
```
rnewson: Our current conversion from JSON text to internal and on-disk representation is order-preserving within objects, which is not strictly necessary. Can we clarify in this RFC if we're intentionally preserving that or not?

rnewson: and a comment on our policy on duplicate keys since we're writing an RFC. "preserve first", "preserve last" or "error if duplicates" seem the only options.

rfcs/004-document-storage.md
```diff
@@ -0,0 +1,246 @@
...
+The disadvantages of this approach are that it uses a larger number of key-value
+pairs and has a higher overall storage overhead from the repeated common key
+prefixes. The new FoundationDB storage engine should eliminate some of the
+storage overhead.
+As per the FoundationDB discussion about being able to co-locate compute operations with data storage servers/nodes](https://forums.foundationdb.org/t/feature-request-predicate-pushdown/954/6), if we were to make use of this hypothetical feature, we’d not get a guarantee of entire documents being co-located on one storage node, requiring us to do extra work should we want to, say, assemble a full `doc` to send to a map function. JS views would have a harder time, while Mango indexes with their explicit field declarations might get around this particular complexity more easily. For now, this is recorded here so we don’t forget track of this later.
+
+
+# Key Changes
+
+- Individual strings within documents are limited to 100 KB each.
```
rnewson: KiB.

kocolosk: Are you sure? From https://apple.github.io/foundationdb/known-limitations.html:

> Keys cannot exceed 10,000 bytes in size. Values cannot exceed 100,000 bytes in size. Errors will be raised by the client if these limits are exceeded.

rnewson: aah, ok, then it's later when you say `MiB` that should be `MB` ("Transaction size cannot exceed 10,000,000 bytes of affected data.").

rnewson: hm, not even that? this is confusing. All the FDB limits are base 10, but the 1 MiB limit is CouchDB's and base 2. Confusing.

rfcs/004-document-storage.md
kocolosk: Good catch. When reading https://apple.github.io/foundationdb/data-modeling.html#encoding-data-types I get the sense that FDB primarily intends the tuple layer to be used with keys, not values, since much of the encoding work is concerned with getting well-defined orderings of the encoded binaries. For example

> For values, the main concerns for serialization are simply CPU and space efficiency. For keys, there’s an additional consideration: it’s often important for keys to preserve the order of the data types ...

@davisp - what do you think? Does it make sense to use some alternative faster / more compact encoding of values?

rfcs/004-document-storage.md
kocolosk: Correct. I don't mind changing the proposed CouchDB limit to base10 so we can be consistent in our overall limitations that we advertise.

rfcs/004-document-storage.md
kocolosk: OK will do. Definitely not order-preserving within objects. Regarding duplicates, probably makes sense tol double-check what Jiffy's Maps-based implementation does and propose the same here /cc @davisp

rfcs/004-document-storage.md
rnewson: Let's change the CouchDB limit to base10 then.

rfcs/004-document-storage.md
davisp: @kocolosk In general I'd agree that the tuple layer is largely more focused on keys, however its also a language agnostic encoding which allows tooling written in other languages to easily interoperate. I haven't got the slightest if that's something we'd want to enable or perhaps even actively discourage.

rfcs/004-document-storage.md
davisp: Jiffy's dedupe option keeps the last key seen as that seemed to be the industry standard.

rfcs/004-document-storage.md
davisp: Also I should point out that the top level tuple is elided in the tuple layer. So `pack({123})` ends up producing the bytes that were most likely assumed would be produced by `pack(123)`. There's no length header when unpacking or anything. It just reads values until the binary is exhausted and then converts the list to a tuple. This makes it nice so that you can "extend" previously encoded tuples and the like (i.e., the subspace layer).

davisp:
Does the suggested 1MiB limit and associated conversation imply that we'll only ever update a single document in a given transaction? One thing I noodled over for a bit for `_bulk_docs` was whether we'd provide single transaction semantics or if each doc update would be an independent transaction.

kocolosk:
@davisp no I did not mean to imply that multi-doc transactions are off the table. I figured it was a separate topic though.

rfcs/004-document-storage.md
```diff
@@ -0,0 +1,246 @@
...
+Document reads where we already know the specific revision of interest can be
+done efficiently using a single `get_range_startswith` operation. In the more
+common case where we do not know the revision identifier, there are two basic
+options:
+
+1. We can retrieve the winning revision ID from the `?REVISIONS` subspace, then
+   execute a `get_range_startswith` operation as above.
+1. We can start streaming the entire key range from the `?DOCUMENTS` space
+   prefixed by `DocID` in reverse, and break if we reach another revision of the
+   document ID besides the winning one.
```
davisp: A couple things to note here. Generally speaking, when we know a specific revision we want to read we don't actually know if its deleted or not so we would have to do two range reads for the `deleted=true` and `deleted=false`. Obviously one of those will be empty but its a slight complication on the stated approach.

For the two options on reading a doc that's obviously up in the air if the extra round trip against the `?REVISIONS` subspace vs possibly streaming extra rows vs possibly having to issue a second range read. Also if we go with Option 1 we don't need to NotDeleted at all which also simplifies things a bit if we're always reading this range with a known `{RevPos, RevHash}` pair.

Another happy side effect of always reading the winning revision from the `?REVISIONS` subspace is that we stop having to worry about deleted vs normal revisions in this subspace. An empty doc with no attachments could just as well have zero data in this space and then we don't have to store something arbitrarily in the metadata for each empty doc as a third kind of odd document state.

rfcs/004-document-storage.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Document Subspace and Versioning
+
+Document bodies will be stored in their own portion of the keyspace with a fixed
+single-byte prefix identifying the "subspace". Each revision of a document will
+be stored separately without term sharing, and the document ID and revision ID
+are baked into the key. The structure looks like this
+
+```
+{DbName, ?DOCUMENTS, DocID, NotDeleted, RevPos, RevHash} = RevisionMetadata
```
davisp: It occurs to me that this format prevents RevisionMetadata from every exceeding 100KB which depending on what we end up putting in there may or may not be an issue.

There's also no versioning in this RFC for how we might do schema migrations in the future like was done for the revision storage RFC. Its obviously possible to repurpose that version marker as a means to version the doc body storage as well but that seems like something we'd want to call out explicitly in one place or other other and also feels a bit limiting that we're tying the two together.

kocolosk: Versioning is included, see the line directly below this one:

> RevisionMetadata includes at the minimum an enum to enable schema evolution for subsequent changes to the document encoding structure

In fact, that's the *only* planned use for that field at the moment. Seems to me that if we have more than 100KB of metadata here, not including the rev tree, something has gone horribly wrong with our database design.

rfcs/004-document-storage.md
kocolosk: Yes I covered this in the CRUD operations section, though I proposed doing the followup read in `?REVISIONS` rather than issuing both `deleted=true` and `deleted=false` guesses.

> If a reader is implementing Option 2 and does not find any keys associated with the supplied DocID in the ?DOCUMENTS space, it will need to do a followup read on the ?REVISIONS space in order to determine whether the appropriate response is {"not_found": "missing"} or {"not_found": "deleted"}.

rfcs/004-document-storage.md
davisp: Ah. I was definitely making the assumption that this would hold things like the #docs.atts metadata and what not. We can obviously do other things there but it still feels like something to make sure that we call out so we ensure that we don’t allow a pathological breakage there.

rfcs/004-document-storage.md
davisp: Ah. Yet again I was slightly off track considering when we return the deleted body which is called out elsewhere for requiring extra fdb requests.

That seems legit but I am starting to worry a bit about the subtlety of some of these optimizations that presime existing behaviors as opposed to being less efficient but easier to implement.

kocolosk: I'm definitely not trying to be "cute" or overly subtle, so stop me if you think there's a better path. My thinking on using `?REVISIONS` in a followup request was exactly because of this condition you brought up:

> An empty doc with no attachments could just as well have zero data in this space

rfcs/004-document-storage.md
kocolosk: Good clarification, thanks. I think we want a value encoding which delivers a) interoperability, b) CPU efficiency and c) storage efficiency. If there are other encodings which deliver better on those traits than the Tuple encoding it could make good sense to use them.

rfcs/004-document-storage.md
davisp: I don't think this is quite to that level. I'm just trying to contemplate if we go with the first option and always require a read from the `?REVISIONS` key space then there's a good chunk of things that end up falling away complexity wise. We'd have defined bounds on all reads to the `?DOCUMENTS` key space and our range reads would be optimal.

However, same as you I believe, forcing that extra round trip for every document read is tempting to try and eliminate. However, for the single range read to get the current winner I'm also wondering how often that ends up with either a second range request or wasted rows read and how that cost compares to the extra round trip.

Luckily though it should be easy to implement both approaches to compare apples to apples and get a better idea of the relative cost either way.

rfcs/004-document-storage.md
kocolosk: Done.

rfcs/004-document-storage.md
kocolosk: Updated the document to be explicit about both these bits.

rfcs/004-document-storage.md
kocolosk: For now I updated the example to `pack({123})`

rfcs/004-document-storage.md
kocolosk: Agreed this is an area where an A/B test is doable without too much surgery.

rfcs/004-document-storage.md
davisp: One idle thought, it seems like we could fairly easily combine things and do something like snappy encoded tuple layer packed binaries. Again its one more thing that can be easily tested to compare. In the meantime I don't think it really affects the rest of the RFC logic so deferring until later doesn't seem unwise.

kocolosk: Yes that's exactly the sort of thing I had in mind. It's slightly less easy to experiment because we're changing the on-disk format, but that's why we have versioning of the data structure.

As I'm sure you've noticed this data model never uses a multi-element tuple for user-supplied data in the values, so any encoding which covers true/false/null, numbers, and strings would suffice on its own.

rfcs/004-document-storage.md
```diff
@@ -0,0 +1,251 @@
...
+{
+    "states": ["MA", "OH", "TX", "NM", "PA"]
+}
+```
+
+becomes
+
+```
+pack({"states", 0}) = "MA"
+pack({"states", 1}) = "OH"
```
garrensmith: Is there a specific reason why we don't pack the values for the array?

kocolosk: Nope, just me being sloppy. Good catch.

rnewson:
Is there more to do here?

kocolosk:
Personally I'm still wondering a bit about the value encoding. For example, should we be pursuing a compression option for string values?

rnewson:
The documented tuple format, while motivated by sorting concerns, does look fairly compact, so do you have concerns about using it for values other than, perhaps, that long string values could be compressed? Short strings won't likely compress by much though if we exclude the various header bytes when storing it might be simplest to compress all strings.

PR #410, opened 2019-05-13
add rfc for map indexes

garrensmith:
## Overview

RFC for map indexes

## Checklist


rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+* `keys` are the emitted keys from the map function
+* `values` is the emitted value from the map function
+
+### Key ordering
+
+FoundationDB orders key by byte value which is not how CouchDB currently orders keys. To maintain the way CouchDB currently does view collation, a type value will need to be prepended to each key so that the correct sort order of null < boolean < numbers < strings < arrays < objects is maintained.
+
+Strings will need an additional change in terms of how they are compared with ICU. An ICU sort string will be generated upfront and added to the string key. This value will be used to sort the string in FDB. The original string will be stored so that it can be used when returning the keys to the user.
+
+CouchDB allows duplicate keys to be emitted for an index, to allow for that a counter value will be added to the end of the keys.
```
davisp: Seems like we should add this to the data model rather than as an aside mention here.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Emitting document
+
+In a map function it is possible to emit the full document as the value, this will cause an issue if the document size is larger than FDB’s value limit of 100 KB. We can handle this in two possible ways.  The first is to keep the hard limit of only allowing 100 KB value to be emitted, so if a document exceeds that CouchDB will return an error. This is the preferred option.
+
+The second option is to detect that a map function is emitting the full document and then add in a foreign key reference back to the document subspace. The issue here is that CouchDB would only be able to return the latest version of the document, which would cause consistency issues when combined with the `update=false` argument.
```
davisp: A third option is to just chunk large values across multiple keys. If we're willing to do multiple keys for multiple values then I'm not sure why we wouldn't be willing here. Then again even then we'll still have limits due to the transaction size.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Index Management
+
+For every document that needs to be processed for an index, we have to run the document through the javascript query server to get the emitted keys and values. This means that it won’t be possible to update a map/reduce index in the same transaction that a document is updated. To account for this, we will need to keep an `id index` similar to the `id tree`  that is currently keep in CouchDB. This index will hold the document id as the key and the value would be the keys that were emitted. CouchDB will use this information to know which fields need to be updated, added or removed from the index when a document is changed.  A data model for this would be:
```
davisp: As mentioned on the mailing list, as written here it introduces a new constraint that the sum of all keys must be less than 100KiB. Having a row per key seems reasonable enough though.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+
+### View clean up
+
+When a design document is changed, new indexes will be built and grouped under a new  `View Signature` . The old map indexes will still be in FDB.  CouchDB will need to run a clean up process that scans all the `View Signature`s and links them to a design document. If a `View Signature` is not linked to any design doc, that `View Signature` namespace can be removed. This will be done via the background jobs scheduler.
```
davisp: As written this would be fairly inefficient as we'd have to scan all of the index content to get the list of view signatures that currently exist since there's no concise listing.

A simple approach to avoid that can also remove the need for the background cleaning process completely. If we were to store an pair of indexes for view signatures of the form:

`(<database>, ?VIEW_DDOC_IDS, DesignDocId) = ViewSignature`
`(<database>, ?VIEW_SIGS, ViewSignature) = Counter`

When a design document is created, we store the view signature and bump the view signature counter. One update or deletion we'd get the old signature and decrement the counter. And on 0 we'd clear out the old view index.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Stale = “ok” and stable = true
+
+ With the consistency guarantee’s CouchDB will get from FDB,  stable = true will no longer be an option that CouchDB would support. Similar `stale = “ok”` would now be translated to `update = false`
```
davisp: Are you suggesting that these become no-ops, or would we reject them?

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Size limits
+
+* Emitted keys will not be able to exceed 10 KB
+* Values cannot exceed 100 KB
+* There could be rare cases where the number of key-value pairs emitted for a map function could lead to a transaction either exceeding 10 MB in size which isn’t allowed or exceeding 5 MB which impacts the performance of the cluster. Ideally CouchDB will need to detect these situations and split the transaction into smaller transactions
```
davisp: Given that these are the underlying fdb limits we may want to consider dropping the actual exposed limits down. Or else we'll end up with customers complaining that 10KiB - 1byte doesn't work because we reserve a variable number of bytes for keys and I could see similar things if we end up structuring keys.

Also, as we've seen before, actually counting how large a value is depends on a number of things which makes exact counts a bit fuzzy.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Advantages
+
+* Map indexes will work on FoundationDB with same behaviour as current CouchDB 2.x
```
davisp: I assume you meant CouchDB 1.x here? Given how the `stale` and `stable` parameters change behavior?

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+* Options like stale = “ok” and ‘stable = true’ will no longer be needed
+
+## Disadvantages
+
+* Size limits on key and values
```
davisp: Also no reduce functions would probably be considered another disadvantage. Or at least noted as a change.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+
+* couch_mrview will be removed and replaced with a new indexing OTP application
+
+## HTTP API additions
+
+The API will remain the same.
+
+## HTTP API deprecations
+
+No deprecations.
```
davisp: `stale` and `stable` were mentioned a couple times. Something somehow would be "deprecated" even if its "You'll not get possibly weird data if you elect to do `stale=ok` or so on.

Also reduce functions are something we'll want to call out.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+
+The second option is to detect that a map function is emitting the full document and then add in a foreign key reference back to the document subspace. The issue here is that CouchDB would only be able to return the latest version of the document, which would cause consistency issues when combined with the `update=false` argument.
+
+### Index Management
+
+For every document that needs to be processed for an index, we have to run the document through the javascript query server to get the emitted keys and values. This means that it won’t be possible to update a map/reduce index in the same transaction that a document is updated. To account for this, we will need to keep an `id index` similar to the `id tree`  that is currently keep in CouchDB. This index will hold the document id as the key and the value would be the keys that were emitted. CouchDB will use this information to know which fields need to be updated, added or removed from the index when a document is changed.  A data model for this would be:
+
+{?DATABASE, ?VIEWS, <view_signature>, ?VIEWS, ?ID_INDEX, <_id>, <view_id>} -> [emitted keys]
```
davisp: Similar to above, I'd replace the `?VIEWS, ?ID_INDEX` with something like `?VIEW_ID_INDEX`.

rfcs/008-map-indexes.md
davisp: > Ideally CouchDB will need to detect these situations and split the transaction into smaller transactions

Just a heads up that going this route would be super duper complicated make work correctly. The update_seq tracking would end up getting a lot more complicated as each individual committed transaction will have to record its progress through what was originally designed to be a single transaction.

This also brings up the parallel view building mentioned (while not proposed for V1) would also require some what to ensure that all of the parallel workers are recording their progress in such a way that they could be properly restarted in such a way that would not leave an inconsistent index.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+* `?VIEWS` is the standard views namespace.
+* `view_signature` is the design documents `View Signature`
+* `view_id` name of a view defined in the design document
+* `?MAP` is the standard map namespace
+* `keys` are the emitted keys from the map function
+* `values` is the emitted value from the map function
+
+### Key ordering
+
+FoundationDB orders key by byte value which is not how CouchDB currently orders keys. To maintain the way CouchDB currently does view collation, a type value will need to be prepended to each key so that the correct sort order of null < boolean < numbers < strings < arrays < objects is maintained.
```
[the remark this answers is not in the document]
garrensmith: Unfortunately, sext matches to erlang ordering and we don't quite follow the erlang ordering since our atoms are before the numbers.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+```
+
+Where each field is defined as:
+
+* `<database>` is the specific database namespace
+* `?VIEWS` is the standard views namespace.
+* `view_signature` is the design documents `View Signature`
+* `view_id` name of a view defined in the design document
+* `?MAP` is the standard map namespace
+* `keys` are the emitted keys from the map function
```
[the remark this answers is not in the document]
garrensmith: Correct. I'm make the changes

rfcs/008-map-indexes.md
garrensmith: They would be no-ops

rfcs/008-map-indexes.md
garrensmith: Nice idea. I've added it to the rfc.

rfcs/008-map-indexes.md
garrensmith: Good points. I've added a bit more information around those limit. I think we need to do a little experimentation before defining what a user's limit should actually be.

rfcs/008-map-indexes.md
garrensmith: I've added in reduce functions. But we will have an RFC for reduce functions later.

garrensmith:
> I'd also like to spend some more time considering if we might not want to change how view groups work. For instance, changing the view signature to represent a single index (rather than all indexes on a given design doc) would likely be an interesting improvement. The historical reason for grouping indexes was to optimize disk usage of shared B+tree's on disk for the id tree along with also ensuring that two separate indexes are always updated to the same point in time.

I still quite like the idea of keeping the view signature mapped to a design document. An idea that @rnewson and I had was to allow CouchDB to support automatic switching of a design doc from using the old view signature index to the new one only once the new index built. I'll send an email about it to the dev email.

garrensmith:
> The last thing that comes to mind is that the reliance the background workers internals seems a bit hand wavy. I wonder if we couldn't define a more simplistic approach and then apply the non-in-v1 sticker to background workers. Something like:

> (<database>, ?VIEWS, ?BUILDS, ?STATUS, ViewSignature) = <json props like _active_tasks>
(<database>, ?VIEWS, ?BUILDS, ?WATCH) = uuid

> Every HTTP node would then take a watch on the ?WATCH key and re-scan the ?STATUS key and inform HTTP clients waiting on view builds.

I really like this, and I think we will need something similar even with workers to make sure multiple workers don't build the same index. I'm going to give it some thought and then update the rfc with this.

garrensmith:
@davisp thats for a great review. I've updated the rfc based on all your feedback. I've also added a future improvements section to keep track of the ideas that we could do later.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,159 @@
...
+* `values` is the emitted value from the map function
+
+### Key ordering
+
+FoundationDB orders key by byte value which is not how CouchDB currently orders keys. To maintain the way CouchDB currently does view collation, a type value will need to be prepended to each key so that the correct sort order of null < boolean < numbers < strings < arrays < objects is maintained.
+
+Strings will need an additional change in terms of how they are compared with ICU. An ICU sort string will be generated upfront and added to the string key. This value will be used to sort the string in FDB. The original string will be stored so that it can be used when returning the keys to the user.
```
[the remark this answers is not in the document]
kocolosk: The discussion on primary, secondary, and tertiary weights, as well as the compression, leads me to believe that even 7bit clean strings would have rather different sort keys -- but I haven't tested it.

http://userguide.icu-project.org/collation/architecture#TOC-Sort-Keys

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,181 @@
...
+* `view_signature` is the design documents `View Signature`
+* `view_id` name of a view defined in the design document
+* `?MAP` is the standard map namespace
+* `key` is the emitted row key from a map function
+* `count` is a value to allow duplicate keys to be emitted for a document
+* `values` is the emitted value from the map function
+
+### Key ordering
+
+FoundationDB orders key by byte value which is not how CouchDB currently orders keys. To maintain the way CouchDB currently does view collation, a type value will need to be prepended to each key so that the correct sort order of null < boolean < numbers < strings < arrays < objects is maintained.
```
kocolosk: In addition to the type ordering, there's a consideration about how Numbers are represented when used as keys. FoundationDB has a smart encoding for arbitrary-precision integers, floats, doubles, and even arbitrary-precision decimals, each with it's own type code. CouchDB muddles all that together and sorts them all numerically, regardless of the underlying binary representation.

Given that we're talking about data that did a roundtrip through Javascript, do we just coerce all numbers to be double-precision floating point when used as keys?

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,181 @@
...
+
+Strings will need an additional change in terms of how they are compared with ICU. An ICU sort string will be generated upfront and added to the string key. This value will be used to sort the string in FDB. The original string will be stored so that it can be used when returning the keys to the user.
+
+CouchDB allows duplicate keys to be emitted for an index, to allow for that a counter value `count` will be added to the end of the keys.
```
kocolosk: Is there a reason why the counter comes before the `_id`? I think CouchDB would return all the emits for a given (key, id) together, while this sorting would put all the duplicate emits at the end.

rfcs/008-map-indexes.md
kocolosk: A row per key is pretty wasteful -- in most cases `<database>` + `<view_signature>` + `<_id>` will be quite a bit larger than the key itself. I'd rather shove the keys list in this index into the smallest number of KVs possible if someone decides to emit more than 100KB of keys in a single view.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,181 @@
...
+
+`(<database>, ?VIEW_DDOC_IDS, <design_doc_id>) = ViewSignature`
+`(<database>, ?VIEW_SIGS, <view_signature>) = Counter`
+
+When a design document is created, CouchDB will store the design document id and the view signature and set the view signature counter to one. On update or deletion of a design document, CouchDB will get the old signature and decrement the counter. 
+If the counter is 0, then CouchDB will remove the old view index.
+
+### Stale = “ok” and stable = true
+
+ With the consistency guarantee’s CouchDB will get from FDB, `stable = true` will no longer be an option that CouchDB would support and so the argument would be ignored. Similar `stale = “ok”` would now be translated to `update = false`.
```
kocolosk: Out of scope, but as an aside it'd be excellent to offer a "bounded staleness" option; i.e., "serve the view stale if it's missing no more than <N> updates, or is lagging by no more than <M> seconds". We don't have the bookkeeping to offer that at the moment, just planting the idea here in case anyone thinks of an elegant way to track that.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,181 @@
...
+
+### View clean up
+
+When a design document is changed, new indexes will be built and grouped under a new `View Signature`. The old map indexes will still be in FDB.  CouchDB will need to monitor `View Signatures` and be able to remove old indexes. To do this we create the following data model:
+
+`(<database>, ?VIEW_DDOC_IDS, <design_doc_id>) = ViewSignature`
+`(<database>, ?VIEW_SIGS, <view_signature>) = Counter`
+
+When a design document is created, CouchDB will store the design document id and the view signature and set the view signature counter to one. On update or deletion of a design document, CouchDB will get the old signature and decrement the counter. 
+If the counter is 0, then CouchDB will remove the old view index.
```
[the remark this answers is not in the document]
garrensmith: Good point @willholley I think maybe for now we should just implement that endpoint and later we can look at adding it as a background task.

garrensmith:
@davisp @kocolosk and @willholley I've updated this RFC to be more in line with the current implementation. Would you mind taking a look, it would be nice to merge this in soon.

rfcs/008-map-indexes.md
```diff
@@ -62,15 +62,17 @@ The data model for a map indexed is:
 
 ```
 (<database>, ?DB_VIEWS, <view_signature>, ?VIEW_UPDATE_SEQ) = Sequence
-(<database>, ?DB_VIEWS, <view_signature>, ?VIEW_ID_RANGE, <_id>, <view_id>) = [emitted keys]
+{<database>, ?DB_VIEWS, <view_signature>, ?VIEW_ID_INFO, view_id, ?VIEW_ROW_COUNT} = <row_count>
+{<database>, ?DB_VIEWS, <view_signature>, ?VIEW_ID_INFO, view_id, ?VIEW_KV_SIZE} = <kv_size>
+(<database>, ?DB_VIEWS, <view_signature>, ?VIEW_ID_RANGE, <_id>, <view_id>) = [total_keys, total_size, unique_keys]
 (<database>, ?DB_VIEWS, <view_signature>, ?VIEW_MAP_RANGE, <view_id>, <key>, <_id>, ?ROW_KEYS <count>) = <emitted_keys>
-(<database>, ?DB_VIEWS, <view_signature>, ?VIEW_MAP_RANGE, <view_id>, <key>, <_id>, ?ROW_VALUE, <count>) = <emitted_value>
+(<database>, ?DB_VIEWS, <view_signature>, ?VIEW_MAP_RANGE, <view_id>, {<key>, <_id>}, <count>, ?ROW_VALUE) = <emitted_value>
```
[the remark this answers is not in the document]
kocolosk: I'd also like to understand the reasoning for using two rows for every emitted KV pair. It's a lot of overhead, particularly with the current FDB storage engine which doesn't have prefix elision. Is it simply aiming to preserve the full 100KB value space for user-emitted values?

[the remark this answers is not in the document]
mikerhodes: I think this is because of:

> When the emitted keys are encoded to binary to create the FDB key for a row, strings are converted to sort strings and all numbers are converted to doubles. It is not possible to convert either of those encoded values back to the original value.

However, I'm inferring; the spec should be clearer.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,202 @@
...
+- `?VIEW_UPDATE_SEQ` is the change sequence namespace
+- `?VIEW_ID_RANGE` is the map id index namespace
+- `?VIEW_MAP_RANGE` is the map namespace
+- `?ROW_KEYS` is the namespace for a row where the value is the emitted keys from a map function
+- `?ROW_VALUE` is the namespace for a row where the value is the emitted value from a map function
+- `_id` is the document id
+- `view_id` id of a view defined in the design document
+- `key` is the encoded emitted row key from a map function
+- `count` is a value that is incremented to allow duplicate keys to be emitted for a document
+- `emitted_key` is the emitted key from the map function
```
mikerhodes: In the data model block, there is no `emitted_key`, only `emitted_keys`. 

I think this difference is important, as if it's `emitted_key` then I don't see how we have the later limit that "The sum of all keys emitted for a document cannot exceed 100 KB" as something that comes from FoundationDB.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,202 @@
...
+1. FDB Transaction is started
+1. Read the document from the changes read (The number of documents to read at one type is configurable, the default is 100)
+1. The document is passed to the javascript query server and run through all the map functions defined in the design document
+1. The view's sequence number is updated to the sequence the document is in the changes feed.
+1. The emitted keys are stored in the `?VIEW_ID_RANGE`
+1. The emitted keys are encoded then added to the `?VIEW_MAP_RANGE` with the emitted keys and value stored
+1. The `?VIEW_ROW_COUNT` is incremented
+1. The `?VIEW_KV_SIZE` is increased
+1. If the document was deleted and was previously in the view, the previous keys for the document are read from `?VIEW_ID_RANGE` and then cleared from the `?VIEW_MAP_RANGE`. The Row count and size count are also decreased.
+1. If the document is being updated and was previously added to the index, then he previous keys for the document are read from `?VIEW_ID_RANGE` and then cleared from the `?VIEW_MAP_RANGE` and then the index is updated with the latest emitted keys and value.
```
mikerhodes: In these steps, be specific about what is being carried out in a single FoundationDB transaction. It sounds like everything including and following "The view's sequence number is updated to the sequence the document is in the changes feed." has to be in the same transaction to be correct.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Key ordering
+
+FoundationDB orders key by byte value which is not how CouchDB orders keys. To maintain CouchDB's view collation, a type value will need to be prepended to each key so that the correct sort order of null < boolean < numbers < strings < arrays < objects is maintained.
+
+In CouchDB 2.x, strings are compared via ICU. The way to do this with FoundationDB is that for every string an ICU sort string will be generated upfront and added to the key.
```
mikerhodes: By "added to the key" do you literally mean that the key will include the byte string and the ICU sort string byte string (this is what I read from "added") or that the key used within FoundationDB for the emitted key will only include the ICU sort string byte string?

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,202 @@
...
+
+A third option would be to split the document across multiple keys in FoundationDB. This will still be limited by the transaction size limit.
+
+### Index building
+
+An index will be built and updated via a [background job worker](https://github.com/apache/couchdb-documentation/blob/master/rfcs/007-background-jobs.md). When a request for a view is received, the request process will add a job item onto the background queue for the index to be updated. A worker will take the item off the queue and update the index. Once the index has been built, the background job server will notify the request that the index is up to date. The request process will then read from the index and return the results. This process can also be optimised in two ways. Firstly, using a new couch_events system to listen for document changes in a database and then adding indexing jobs to the queue to keep indexes warm. The second optimisation is if the index only requires a small update, rather update the index in the HTTP request process instead of doing the work via the background queue.
```
mikerhodes: I'm a bit dubious about adding indexing workloads to HTTP workers if the deployment decides to split out different workers for HTTP and indexing. Is there an alternative via creating some kind of "fast path" for couch_jobs where the HTTP worker is able to directly contact an index worker to get it to perform the work rather than work via FoundationDB?

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,202 @@
...
+
+In CouchDB 2.x, strings are compared via ICU. The way to do this with FoundationDB is that for every string an ICU sort string will be generated upfront and added to the key.
+
+### Emitting document
+
+In a map function, it is possible to emit the full document as the value, this will cause an issue if the document size is larger than FDB’s value limit of 100 KB. We can handle this in two possible ways. The first is to keep the hard limit of only allowing 100 KB value to be emitted, so if a document exceeds that CouchDB will return an error. This is the preferred option.
+
+The second option is to detect that a map function is emitting the full document and then add in a foreign key reference back to the document subspace. The issue here is that CouchDB would only be able to return the latest version of the document, which would cause consistency issues when combined with the `update=false` argument.
+
+A third option would be to split the document across multiple keys in FoundationDB. This will still be limited by the transaction size limit.
```
mikerhodes: Can this section be collapsed now into "the approach that has been chosen and why" and "the approaches that were not chosen and why".

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,202 @@
...
+(<database>, ?DB_VIEWS, <view_signature>, ?VIEW_ID_RANGE, <_id>, <view_id>) = [total_keys, total_size, unique_keys]
+(<database>, ?DB_VIEWS, <view_signature>, ?VIEW_MAP_RANGE, <view_id>, <key>, <_id>, ?ROW_KEYS <count>) = <emitted_keys>
+(<database>, ?DB_VIEWS, <view_signature>, ?VIEW_MAP_RANGE, <view_id>, {<key>, <_id>}, <count>, ?ROW_VALUE) = <emitted_value>
+```
+
+Each field is defined as:
+
+- `<database>` is the specific database namespace
+- `?DB_VIEWS` is the views namespace.
+- `view_signature` is the design documents `View Signature`
```
mikerhodes: It'd be worth specifying whether this will be the same value as calculated for view sig by CouchDB 2.x. 

It's possible to imagine that, if it did, a migration path could stream direct from CouchDB 2.x index files into FoundationDB, so I think it's important to specify one way or the other.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,202 @@
...
+- `view_id` id of a view defined in the design document
+- `key` is the encoded emitted row key from a map function
+- `count` is a value that is incremented to allow duplicate keys to be emitted for a document
+- `emitted_key` is the emitted key from the map function
+- `emitted_value` is the emitted value from the map function
+- `row_count` number of rows in the index
+- `kv_size>` size of the index
+- `total_keys` is the number of keys emitted by a document
+- `total_size` is the size of the key/values emitted by the document
+- `unique_keys` is the unique keys emitted by the document
```
willholley: is this the value of the unique keys or the count?

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,202 @@
...
+### Data model
+
+The data model for a map indexed is:
+
+```
+(<database>, ?DB_VIEWS, <view_signature>, ?VIEW_UPDATE_SEQ) = Sequence
+{<database>, ?DB_VIEWS, <view_signature>, ?VIEW_ID_INFO, view_id, ?VIEW_ROW_COUNT} = <row_count>
+{<database>, ?DB_VIEWS, <view_signature>, ?VIEW_ID_INFO, view_id, ?VIEW_KV_SIZE} = <kv_size>
+(<database>, ?DB_VIEWS, <view_signature>, ?VIEW_ID_RANGE, <_id>, <view_id>) = [total_keys, total_size, unique_keys]
+(<database>, ?DB_VIEWS, <view_signature>, ?VIEW_MAP_RANGE, <view_id>, <key>, <_id>, ?ROW_KEYS <count>) = <emitted_keys>
```
mikerhodes: Is the `count` in this row the same as the `count` in the following row?

rfcs/008-map-indexes.md
garrensmith: It should be `emitted_keys`

rfcs/008-map-indexes.md
garrensmith: Its a list of all the unqiue keys

mikerhodes:
@garrensmith For me this would be much easier to follow with a worked example. As we're still dealing in view groups here, I suggest something like:

Two docs:

```
{
    "_id": "foo",
    "hello": "mike"
}
{
    "_id": "bar",
    "hello": "garren"
}
```

Two map indexes:

```
// view name "numbers"
function map(doc) {
    emit([1,2,3], doc.hello);
}

// view name "letters"
function map(doc) {
    emit("b", doc.hello);
}
```

rfcs/008-map-indexes.md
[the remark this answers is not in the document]
kocolosk: The sort key transformation is the reason we need to store `<emitted_key>` at all, but I'm conending that a single FDB Value could store both the `<emitted_key>` and `<emitted_value>`.

rfcs/008-map-indexes.md
```diff
@@ -0,0 +1,202 @@
...
+- Emitted keys will not be able to exceed 10 KB
+- Values cannot exceed 100 KB
+- There could be rare cases where the number of key-value pairs emitted for a map function could lead to a transaction either exceeding 10 MB in size which isn’t allowed or exceeding 5 MB which impacts the performance of the cluster. In this situation, CouchDB will send an error.
+
+These limits are the hard limits imposed by FoundationDB. We will have to set the user imposed limits to lower than that as we store more information than just the user keys and values.
+
+## Advantages
+
+- Map indexes will work on FoundationDB with the same behaviour as current CouchDB 1.x
+- Options like stale = “ok” and ‘stable = true’ will no longer be needed
```
kocolosk: Agreed on `stable=true` but `stale=ok` is still a meaningful option in this design.

rfcs/008-map-indexes.md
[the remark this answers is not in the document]
garrensmith: I'm happy to move to a single row that contains `<emitted_key> and <emitted_value>` as long as we ok with the new limits:

* Emitted key must be smaller than 10Kb
* Emitted value must be smaller than 90Kb.

rfcs/008-map-indexes.md
```diff
@@ -31,17 +31,17 @@ document are to be interpreted as described in
 
 ## Terminology
 
-`Sequence`: a 13 byte value formed by combining the current `Incarnation` of the database and the `Versionstamp` of the transaction. Sequences are monotonically increasing even when a database is relocated across FoundationDB clusters. See (RFC002)[LINK TBD] for a full explanation.
+`Sequence`: a 13-byte value formed by combining the current `Incarnation` of the database and the `Versionstamp` of the transaction. Sequences are monotonically increasing even when a database is relocated across FoundationDB clusters. See (RFC002)[LINK TBD] for a full explanation.
```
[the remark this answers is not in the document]
davisp: Right. It was just copied from the original. Would probably be better to have links out to other RFCs for a centralized definition instead of repeating it everywhere.

rfcs/008-map-indexes.md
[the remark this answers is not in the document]
davisp: @kocolosk The general logic was that previously there was no size limit for key/value sizes so moving to fdb we'd want to preserve as much space as possible to minimize the number of confusing situations where a view on fdb would appear to work but then have a subtle difference on some rows just not appearing without user feedback that there was something amiss.

Originally I hadn't really considered adding size limits to the key/value independently. I was also worried about how to explain to users "The size of the key and value must be less than 100Kib." which was another reason I thought two rows would be better since the documentation would be more straightforward.

rfcs/008-map-indexes.md
davisp: While its currently a direct copy of the logic I'm not sure its a good idea to assert the equivalence. I'm reasonably certain that the implementation doesn't touch on anything that's changed, but I could easily see a subtle change that does alter the calculation.

rfcs/008-map-indexes.md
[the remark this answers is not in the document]
kocolosk: OK. I don't think that's worth doubling the overhead.

I'd like to leave a buffer so that we've got options for things like storing Values in an encrypted form. We also need a buffer on the key for several reasons:

- sort key generation
- database name (are we really using the raw database name here?)
- view signature

Agreed that it would be bad if views silently crash and omit rows -- at the risk of scope creep do we need to include in the data model the list of documents that failed to index that we had discussed in the past?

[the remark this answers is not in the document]
davisp: I'm not against enforcing those limits by any means but we'll want to go through and figure out all of the extra requirements/behaviors and how we communicate those issues to users.

As far as I recall, the tracking documents that failed to be indexed was agreed to in theory but then got hand waved to future version land. I've not got much of an opinion on whether or not its part of this RFC or a later extension. Though I may lean towards later since it's going to have to include user visible API decisions and so on.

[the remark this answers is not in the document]
mikerhodes: Hmm, frankly, at this point we're getting to something like:

- 5KB key
- 50KB value

This leaves a decent chunk of overhead for cryptography. We would need to enforce them in the indexing code itself rather than leaving it to FDB (I assume we are doing that anyway, but this pushes us to really need them).

garrensmith:
I've updated this map index RFC so it is up to date with how it is currently implemented.

PR #414, opened 2019-05-23
Correct "Refuse building with known bad versions of Erlang" issue number

dannyroberts:
## Overview

The docs list a bugfix "Refuse building with known bad versions of Erlang" and incorrectly reference it as issue 1875, when it's actually issue 1857.


wohali:
Thanks for the correction!

PR #415, opened 2019-05-28
Add rfc for using ExUnit for unit tests

iilyak:
## Overview

RFC for using ExUnit for unit tests

## Testing recommendations

N/A

## GitHub issue number

N/A

## Related Pull Requests

- Initial implementation of the RFC https://github.com/apache/couchdb/pull/2039

## Checklist

### DIFF
--- a/rfcs/009-exunit.md
+++ b/rfcs/009-exunit.md
@@ -1,0 +1,122 @@
+---
+name: Formal RFC
+about: Submit a formal Request For Comments for consideration by the team.
+title: 'Use ExUnit testing framework for unit testing'
+labels: rfc, discussion
+assignees: ''
+
+---
+
+# Introduction
+
+With the upgrade of supported Erlang version and introduction of Elixir into our
+integration test suite we have an opportunity to replace currently used eunit
+(for new tests only) with Elixir based ExUnit. 
+
+## Abstract
+
+Eunit testing framework has a number of issues which makes it very hard to use.
+We already use alternative testing framework called ExUnit for integration tests.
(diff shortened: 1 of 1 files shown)


rfcs/009-exunit.md
```diff
@@ -0,0 +1,129 @@
...
+about: Submit a formal Request For Comments for consideration by the team.
+title: 'Use ExUnit testing framework for unit testing'
+labels: rfc, discussion
+assignees: ''
+
+---
+
+# Introduction
+
+With the upgrade of supported Erlang version and introduction of Elixir into our
```
garrensmith: I think just this paragraph should be the introduction. Then the section below should be below the abstract as an actual reason why we want to move to exunit.

rfcs/009-exunit.md
```diff
@@ -0,0 +1,129 @@
...
+- it is hard to skip certain tests when needed
+
+ExUnit shouldn't have these problems:
+- on_exit function is reliable in ExUnit
+- it is easy to group tests using `describe` directive
+- code-generation is trivial, which makes it is possible to generate tests from formal spec (if/when we have one)
+
+## Abstract
+
+Write necessary plumbing to be able to use ExUnit in unit tests.
```
garrensmith: I find the abstracts really confusing but I think this should be more a full summary of the whole RFC. Which is something like, we want to replace eunit because it has a bunch of issues, we will use exunit instead, it solves a lot of these problems. The advantage is we would then use a modern testing framework which is consistent for unit and integration tests.

iilyak:
@garrensmith I moved sections as you've advised.

PR #416, opened 2019-06-06
Mention that _ensure_full_commit is now a no-op

flimzy:
Follow-up to an earlier comment: https://github.com/apache/couchdb/issues/1585#issuecomment-417059810

Will this endpoint ever be revived in the future? If not, perhaps it's worth removing some of the details, and even updating other mentions of it (particularly in the replication protocol documentation)?


janl:
this was done in https://github.com/apache/couchdb-documentation/pull/445, sorry for the dupe

Issue #417, opened 2019-06-07
Update documentation to describe using search indexes


lbboe:
Search index needs to be documented.

kocolosk:
Closed by #418

PR #418, opened 2019-06-07
Add search index documentation

lbboe:
## Overview

Add documentation about using search indexes to query a database using Lucene Query Parser Syntax. 

## GitHub issue number

https://github.com/apache/couchdb-documentation/issues/417


src/api/ddoc/views.rst
```diff
@@ -315,6 +315,1128 @@ including the update sequence of the database from which the view was
 generated. The returned value can be compared this to the current update
 sequence exposed in the database information (returned by :get:`/{db}`).
 
+Search
```
flimzy: Would a better title be 'Indexes' or 'Search Indexes'?  There are many ways to 'search' in CouchDB, so this seems like an ambiguous title.

src/api/ddoc/views.rst
```diff
@@ -315,6 +315,1128 @@ including the update sequence of the database from which the view was
...
+
+To create a search index, you add a JavaScript function to a design document in the database. An index builds after processing one search request or after the server detects a document update. The ``index`` function takes the following parameters: 
+
+1.  Field name - The name of the field you want to use when you query the index. If you set this parameter to ``default``, then this field is queried if no field is specified in the query syntax.
+2.  Data that you want to index, for example, ``doc.address.country``. 
+3.  (Optional) The third parameter includes the following fields: ``boost``, ``facet``, ``index``, and ``store``. These fields are described in more detail later.   
+
+By default, a search index response returns 25 rows. The number of rows that is returned can be changed by using the ``limit`` parameter. However, a result set from a search is limited to 200 rows. Each response includes a ``bookmark`` field. You can include the value of the ``bookmark`` field in later queries to look through the responses.
+
+You can query the API by using one of the following methods: URI, CouchDB Dashboard, curl, or a browser plug-in, such as Postman or RESTClient.
```
flimzy: Does this sentence belong here? It doesn't seem particularly relevant to search indexes. It seems like general information, which might belong in a "getting started" section, rather than in something about indexes?

src/api/ddoc/views.rst
```diff
@@ -315,6 +315,1128 @@ including the update sequence of the database from which the view was
...
+Index functions
+---------------
+
+Attempting to index by using a data field that does not exist fails. To avoid this problem, use an appropriate :ref:`index_guard_clauses`.
+
+.. note:: 
+    Your indexing functions operate in a memory-constrained environment where the 
+    document itself forms a part of the memory that is used in that environment. 
+    Your code's stack and document must fit inside this memory. Documents are limited 
+    to a maximum size of 64 MB.
```
flimzy: Does this 64mb limit apply only to search index functions? If so, this might need to be clarified. If this limit is system-wide, does it need to be (re-)mentioned here?

Assuming the limit only applies to search index functions, what is the consequence of a > 64mb doc? I gather, from reading below, that it will simply be omitted from the index, but it might be good to clarify.

src/api/ddoc/views.rst
```diff
@@ -315,6 +315,1128 @@ including the update sequence of the database from which the view was
...
++------------------------+------------------------------------------------------+-------------------+------------------+-----------------------+-------------------+
+| ``group_field``        | Field that groups search matches                     | yes               |  String          | A string that         | no                |
+|                        |                                                      |                   |                  | contains the name of  |                   |
+|                        |                                                      |                   |                  | a string field.       |                   |
+|                        |                                                      |                   |                  | Fields containing     |                   |
+|                        |                                                      |                   |                  | other data such as    |                   | 
+|                        |                                                      |                   |                  | numbers,objects, or   |                   |
```
flimzy:
```suggestion
|                        |                                                      |                   |                  | numbers, objects, or  |                   |
```

src/api/ddoc/views.rst
rnewson: since this follows the title 'Querying Views and Indexes', I think 'Search' is about as good as you'll get. This is the full-text search feature which _is_ distinct from colloquial uses of the word 'search'.

src/api/ddoc/views.rst
rnewson: agree that this sentence seems unnecessary.

src/api/ddoc/views.rst
rnewson: this is referring to the (default) 64 MiB document size limit, but the point being made here is that the document has to be loaded in order to be indexed. It's appropriate to mention that here, but we should clarify that the 64 MiB limit can be altered.

src/api/ddoc/views.rst
rnewson: +1

src/api/ddoc/views.rst
wohali: @flimzy also, we have started referring to the set of (Mango, JavaScript views, and Search) all as "secondary indexes", so this is fine.

src/api/ddoc/views.rst
```diff
@@ -315,6 +315,1144 @@ including the update sequence of the database from which the view was
 generated. The returned value can be compared this to the current update
 sequence exposed in the database information (returned by :get:`/{db}`).
 
+Search
+======
+
+Search indexes enable you to query a database by using `Lucene Query Parser Syntax <http://lucene.apache.org/core/4_3_0/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#Overview>`_. A search index uses one, or multiple, fields from your documents. You can use a search index to run queries, find documents based on the content they contain, or work with groups, facets, or geographical searches.
+
```
wohali: I want to see a callout box here saying that Search is non-functional without a functioning, cluster-connected Clouesau instance, so that people don't search our docs for "Search" or "Lucene" and think the functionality is provided out of the box.

Sample callout box syntax is here, below the table (line 52): https://github.com/apache/couchdb-documentation/blob/299cc785bf1e631401d13444741d2fac80788513/src/setup/cluster.rst#ports-and-firewalls

src/api/ddoc/views.rst
```diff
@@ -315,6 +315,1144 @@ including the update sequence of the database from which the view was
...
+Search index partitioning type
+------------------------------
+
+A search index will inherit the partitioning type from the ``options.partitioned``
+field of the design document that contains it.
+
+Index functions
+---------------
+
+Attempting to index by using a data field that does not exist fails. To avoid this problem, use an appropriate :ref:`index_guard_clauses <api/ddoc/view>`.
```
wohali: This line is too long. Our documentation maximum line length is 90 characters. The automated tooling will probably start to complain about this after you fix the table formatting issue.

src/api/ddoc/views.rst
```diff
@@ -315,6 +315,1144 @@ including the update sequence of the database from which the view was
...
+you do not have to specify a field name at query time.
+The effect is that the query can be simplified:
+
+.. code-block:: javascript
+
+    query=red
+
+The second parameter is the data to be indexed. Keep the following information in mind when you index your data: 
+
+- This data must be only a string, number, or boolean. Other types will cause an error to be thrown by the index function call.
```
wohali: More lines that are too long here.

src/api/ddoc/views.rst
```diff
@@ -315,6 +315,1144 @@ including the update sequence of the database from which the view was
...
+- If an error is thrown when running your function, for this reason or others, the document will not be added to that search index.
+
+The third, optional, parameter is a JavaScript object with the following fields:
+
+*Index function (optional parameter)*
+
++--------------+----------------------------------------------------------------------+----------------------------------+-----------------+
```
wohali: If you have trouble getting this table to format correctly, you might try the list-table format:

http://docutils.sourceforge.net/docs/ref/rst/directives.html#list-table

We've not used one of these before, but Sphinx (the tool we use for building the docs) claims it's supported here: http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#table-directives

If this works out better for you, but the build breaks still because our version of Sphinx / Docutils is too old, I'll get that fixed for you.

src/api/ddoc/views.rst
```diff
@@ -315,6 +315,1144 @@ including the update sequence of the database from which the view was
...
+			    "index": "function (doc) { ... }",
+			    "analyzer": "$ANALYZER_NAME"
+		    }
+	    }
+    }
+
+.. _api/ddoc/view/language-specific-analyzers:
+
+Language-specific analyzers
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
wohali: These all come from (a specific version of) Lucene, yes? Perhaps we should include a link back to Lucene's documentation where the full list is also specified.

src/api/ddoc/views.rst
```diff
@@ -315,6 +315,1144 @@ including the update sequence of the database from which the view was
...
+The available logical operators are case-sensitive and are ``AND``, ``+``, ``OR``, ``NOT`` and ``-``.
+Range queries can run over strings or numbers.
+
+If you want a fuzzy search,
+you can run a query with ``~`` to find terms like the search term.
+For instance,
+``look~`` finds the terms ``book`` and ``took``.
+
+.. note::
+    If the lower and upper bounds of a range query are both strings that contain only numeric digits, the bounds are treated as numbers not as strings. For example, if you search by using the query ``mod_date:["20170101" TO "20171231"]``, the results include documents for which ``mod_date`` is between the numeric values 20170101 and 20171231, not between the strings "20170101" and "20171231".
```
wohali: line too long

src/api/ddoc/views.rst
```diff
@@ -315,6 +315,1144 @@ including the update sequence of the database from which the view was
...
+                "expensive": 278682,
+                "cheap": 257023
+            }
+        }
+    }
+
+.. _api/ddoc/view/geographical_searches:
+
+Geographical searches
+---------------------
```
wohali: Do we need a disclaimer here that this is Lucene's implementation, not the fancy geo index Cloudant also has open sourced? It probably only matters to geo-heads, but if I understand correctly, there's no map projection going on in Lucene, this is just a numerical bounding box.

kocolosk: Makes sense, maybe just something subtle like 

> you can also sort your results by their distance from a geographic coordinate using Lucene's builtin geospatial capabilities.

For the typical reader it'll just seem like a bit of extra verbiage, but a geo-head will understand it's not a `libspatialindex` thing.

kocolosk:
I just added some documentation of the config settings for Dreyfus in the main PR that probably makes sense to include over here: https://github.com/apache/couchdb/pull/2037/commits/0d32708d68432b127154757643f28efd20066330

src/api/ddoc/views.rst
```diff
@@ -315,22 +313,1355 @@ including the update sequence of the database from which the view was
...
++----------------+----------------------------------------------------------+
+| ``indonesian`` | org.apache.lucene.analysis.id.IndonesianAnalyzer         |
++----------------+----------------------------------------------------------+
+| ``irish``      | org.apache.lucene.analysis.ga.IrishAnalyzer              |
++----------------+----------------------------------------------------------+
+| ``italian``    | org.apache.lucene.analysis.it.ItalianAnalyzer            |
++----------------+----------------------------------------------------------+
+| ``japanese``   | org.apache.lucene.analysis.ja.JapaneseAnalyzer           |
++----------------+----------------------------------------------------------+
+| ``japanese``   | import org.apache.lucene.analysis.ja.JapaneseTokenizer   |
```
kocolosk: Remove `import` here

src/api/ddoc/views.rst
```diff
@@ -315,22 +313,1355 @@ including the update sequence of the database from which the view was
...
++----------------+----------------------------------------------------------+
+| ``russian``    | org.apache.lucene.analysis.ru.RussianAnalyzer            |
++----------------+----------------------------------------------------------+
+| ``spanish``    | org.apache.lucene.analysis.es.SpanishAnalyzer            |
++----------------+----------------------------------------------------------+
+| ``swedish``    | import org.apache.lucene.analysis.sv.SwedishAnalyzer     |
++----------------+----------------------------------------------------------+
+| ``thai``       | import org.apache.lucene.analysis.th.ThaiAnalyzer        |
++----------------+----------------------------------------------------------+
+| ``turkish``    | import org.apache.lucene.analysis.tr.TurkishAnalyzer     |
```
kocolosk: Drop the `import` for these three as well.

src/api/ddoc/views.rst
kocolosk: Looking at the rendered layout, though, the parent section is `10.5.5. /db/_design/design-doc/_view/view-name`, which is incorrect for search. Should we move the entire search section one level up so that it's a peer to `_info`, `_view` and `_search_analyze`?

/cc @wohali @flimzy @rnewson 
Screen Shot 2019-06-26 at 9 30 31 PM

src/api/ddoc/views.rst
wohali: This still needs resolution.

src/api/ddoc/views.rst
wohali: There are still lines that are too long, like this one.

src/api/ddoc/views.rst
wohali: Still many lines that are too long in this document. Please resolve.

src/api/ddoc/views.rst
wohali: Still needs resolution.

src/api/ddoc/views.rst
wohali: Still needed.

kocolosk:
@Wohali hmm I thought the Clouseau fallout and the analyzer list were resolved. You’re definitely right about the geo one, though.

wohali:
@kocolosk mostly, it's the lines-too-long stuff. I think our automated checker isn't faulting the source on it right now, but invariably, it will start doing so as soon as it hits `master` ;)

kocolosk:
That’s weird. The linter was definitely faulting too-long-lines earlier in the PR, and Lora fixed everything it was complaining about. I’ll take a look

kocolosk:
`make check` runs clean locally, fwiw

src/api/ddoc/views.rst
wohali: The challenge is that this is the API reference section - specific to actual endpoints. If this is general search documentation, perhaps it belongs somewhere else, not in the Design Document API reference.

My gut reaction would be to put most of the content in the `ddocs/views` directory as its own top level instead, saving the changes to the API reference explicitly for any endpoint API changes only...though it's worth leaving *something* behind for Search here, since it does indeed end up in a ddoc in the long run.

Also not sure how you're getting 10.x out of this, over on docs.couchdb.org this is [1.5.4](https://docs.couchdb.org/en/latest/api/ddoc/views.html). :confused:

src/api/ddoc/views.rst
wohali: For clarity, this still needs a link back to Apache Lucene's documentation somewhere.

src/api/ddoc/views.rst
kocolosk: Yep OK, that makes sense. But I'd still argue that the Search documentation should go up a level since you hit the `_search` endpoint and not the `_view` one when you make requests.

wohali: s4eamxv

kocolosk: This is done -- well, actually we put the Search User Guide directly under "Design Documents" rather than as a child of "Guide to Views". We also have full documentation for `_search_analyze`, `_search` and `_search_info` in the right places.

src/api/ddoc/views.rst
kocolosk: It's not obvious to me what the best link would be. We could use the javadoc for `org.apache.lucene.analysis.Analyzer` from which all of these inherit:

https://lucene.apache.org/core/4_6_1/core/org/apache/lucene/analysis/Analyzer.html

We could use the package documentation for Analysis, which goes into a bunch of detail about how to use these analyzers and what they do -- most of which is not at all relevant for CouchDB users:

https://lucene.apache.org/core/4_6_1/core/org/apache/lucene/analysis/package-summary.html

I guess it doesn't really matter much. I just don't want to send people down a rabbit hole, so let's make sure we're linking to reference material as opposed to required reading.

kocolosk:
OK, I think this is ready to merge, all requested changes have been made and I'm personally happy with the layout and organization of the User Guide and the additions to the API Reference. Thanks @lbboe for this contribution and thanks @flimzy and @wohali for the marathon review!

PR #424, opened 2019-07-07
rfc(per-doc-access): first draft

### LINKED ISSUE #1524: Better support for db-per-user
@janl says:
>I don’t know what this will look like, but this is a pattern, and we need to support it better.
>
>One approach could be “virtual dbs” that are backed by a single database, but that’s usually at odds with views, so we could make this an XOR and disable views on these dbs. Since this usually powers client-heavy apps, querying usually happens there anyway.
>
>Another approach would be better / easier cross-db aggregation or querying. There are a few approaches, but nothing really slick.

janl:
Addresses https://github.com/apache/couchdb/issues/1524

### DIFF
--- a/rfcs/010-per-document-access-control.md
+++ b/rfcs/010-per-document-access-control.md
@@ -179,16 +179,16 @@ An access-enabled database behaves like this:
      docs have been created/updated/deleted in between two of their
      docs.
 
-    * this includes all the user’s docs PLUS all non-`_access` design
-        docs, so apps can centrally control design docs going down to
-        satellites.
+   * this includes all the user’s docs PLUS all design docs with
+     `_access: ["_users"]`, so apps can centrally control design docs
+     going down to satellites.
 
 * accessing `_all_docs` gives users the subset of docs they own in `_id`
   order.
 
-   * this includes all the user’s docs PLUS all non-`_access` design
-     docs, so apps can centrally control design docs going down to
-     satellites.
+   * this includes all the user’s docs PLUS all design docs with
+     `_access: ["_users"]`, so apps can centrally control design docs
+     going down to satellites.
 
 * Replication check-points / local docs
 
--- a/rfcs/010-per-document-access-control.md
+++ b/rfcs/010-per-document-access-control.md
@@ -141,7 +141,7 @@ An access-enabled database behaves like this:
   work around this and potentially prefix docs with the username before
   writing/replicating them in.
 
-* _security members are allowed to write design docs, but the have to
+* _security members are allowed to write design docs, but they have to
   have an `_access` field and those design docs with an `_access` field
   are ignored on the server side. Db-admin ddocs get indexes built as
   normal.
--- a/rfcs/010-per-document-access-control.md
+++ b/rfcs/010-per-document-access-control.md
@@ -1,0 +1,397 @@
+---
+name: Per-Document Access Control
+about: Make the db-per-user pattern obsolete.
+title: 'Per-Document Access Control'
+labels: rfc, discussion, access control, security
+assignees: '@janl'
+
+---
+
+# Introduction
+
+Up until now (version 2.3.1), CouchDB could not serve mutually
+untrusting users accessing the same database. If a user has access to
+one document in a database, they have access to all other documents in
+the database. Some restrictions can be added about writing documents
+(designs docs are db-admin only, validate doc update (VDU) functions
+could restrict write access based on the writing user and/or the target
+document). For the remainder of this document, “db-admin” SHALL include
+server admins as well.
+
+## Abstract
+
+This lead to CouchDB developers making use of a pattern called
+db-per-user, where all documents belonging to one user are kept in a
+separate database. This is a decent enough workaround, but has the
+following downsides:
+
+- queries across all databases are not possible. An additional
+  workaround exists where all per-user databases are replicated
+  continuously into a central, admin-only database that can be used for
+  querying the entire data set, but that adds latency and uses
+  significant CPU resources. Successful systems have been built where
+  increased latency could be traded for fewer CPU resources, but
+  overall, this is not an optimal design.
+
+- handling many small databases, say >10000 (depending on hardware) can
+  become a challenge, if most of them are active concurrently. It
+  forces dbs to be set to `q=1`, migrating off `q!=1` requires
+  downtime, 10k bidirectional replications are going to need A LOT of
+  CPU and RAM. sharing documents among two or more users requires the
+  creation of yet more databases.
+
+Per-user document access aims to solve many of the above problems.
+Predominantly, that multiple users can use a single database without
+being able to see each other’s documents. A first iteration is not
+going to solve sharing of documents across multiple users and/or groups.
+
+Goals for this iteration of this feature:
+
+* allow developers to build apps wihtout having to resort to using the
+  db-per-user pattern. Specifically PouchDB applications and CouchDB
+  setups with a central server/cluster and many independent satellite
+  installations with replication should be supported.
+
+Non-goals for now:
+
+* per-access views
+* differentiation between read and write access for documents
+* sharing infividual documents between multitple users or groups.
+
+However, the design of this iteration aims to allow turning these
+non-goals into actual goals later.
+
+## Dramatis Personae
+
+*user*: a CouchDB-user, a record defined in the _users db identified by
+a username and password, has associated roles.
+
+*developer*: creator of an application built on top of CouchDB
+
+## Requirements Language
+
+[NOTE]: # ( Do not alter the section below. Follow its instructions. )
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.txt).
+
+---
+
+# Detailed Description
+
+You will be able to create databases with the “access” feature enabled
+via an option passed at database creation time. If you create a
+database without that option, it works like any database in CouchDB
+today.
+
+This is how you create an access-enabled database:
+
+```
+PUT /database?access=true
+```
+
+This option can be set only at database creation time, it can’t be
+turned off and on while the database exists.
+
+An access-enabled database behaves like this:
+
+* only admin users can read or write to the database (as per 3.x
+  defaults)
+
+* admins can grant individual users and groups access to a database
+  using the database’s `_security` object. A special new role `_users`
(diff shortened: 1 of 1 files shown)


ermouth:
Are `skip` and `limit` query params expected to work as in existing versions?

janl:
@ermouth could you give an example?

ermouth:
@janl I mean a protected bucket may have interleaving docs of users A and B, like:
```
{_id: '01', _access:['A']}, {_id:'02', _access:['B']}, {_id: '03', _access:['A']}
```
What exactly `_all_docs?limit=1&skip=1` will return for user A?

janl:
@ermouth empty result like if `_id 02` didn’t exist. `by-access-id` is all docs in order of `_access` + `_id` and then any one authenticated users only gets their `_access` range, but can use that like regular `by-id` (`_all_docs`)

rfcs/010-per-document-access-control.md
```diff
@@ -0,0 +1,416 @@
...
+one document in a database, they have access to all other documents in
+the database. Some restrictions can be added about writing documents
+(designs docs are db-admin only, validate doc update (VDU) functions
+could restrict write access based on the writing user and/or the target
+document). For the remainder of this document, “db-admin” SHALL include
+server admins as well.
+
+## Abstract
+
+This lead to CouchDB developers making use of a pattern called
```
jaydoane: Do you want "led" here as the [past tense of lead](https://www.grammarly.com/blog/led-lead/)?

janl: I do! thx

rfcs/010-per-document-access-control.md
```diff
@@ -0,0 +1,416 @@
...
+        with separate _access entries, it becomes admin-only by
+        default. This case needs to be handled by an applications
+        _conflict handler.
+
+* document _ids are shared across all users. So only the first user who
+  creates the doc `_id: config` gets it. Applications need to ensure to
+  work around this and potentially prefix docs with the username before
+  writing/replicating them in.
+
+* _security members are allowed to write design docs, but the have to
```
janl:
```suggestion
* _security members are allowed to write design docs, but they have to
```

rfcs/010-per-document-access-control.md
```diff
@@ -0,0 +1,416 @@
...
+
+* If an existing doc changes the user mentioned in `_access` or an admin
+  user adds a non-admin user after updating the document a couple of
+  times, that new user will gain access to the full history of the
+  document.
+
+   * if compaction hasn’t run yet, they get access to all previous
+     revision bodies that still exist.
+
+   * all conflicted versions will also be visible to the new user
```
janl:
```suggestion
   * all conflicted versions will also be visible to the new user, unless there are conflicting usernames in conflict’s `_access` property.
```

rfcs/010-per-document-access-control.md
```diff
@@ -0,0 +1,416 @@
...
+     to by trying to recreate an old hash.
+
+* accessing `_changes` gives users the subset of docs they own in last
+  updated order
+
+   * gaps in the sequence id would allow folks to deduce how many other
+     docs have been created/updated/deleted in between two of their
+     docs.
+
+      * this includes all the user’s docs PLUS all non-`_access` design
```
janl:
```suggestion
    * this includes all the user’s docs PLUS all non-`_access` design
```

rfcs/010-per-document-access-control.md
```diff
@@ -0,0 +1,418 @@
...
+title: 'Per-Document Access Control'
+labels: rfc, discussion, access control, security
+assignees: '@janl'
+
+---
+
+# Introduction
+
+Up until now (version 2.3.1), CouchDB could not serve mutually
+untrusting users accessing the same database. If a user has access to
```
garrensmith: The introduction and Abstract don't feel right to me. What you have in the introduction seems to continue in the Abstract. I actually think all of that should go in the detailed description of what is the problem and how we trying to solve it.

rfcs/010-per-document-access-control.md
```diff
@@ -0,0 +1,418 @@
...
+  overall, this is not an optimal design.
+
+- handling many small databases, say >10000 (depending on hardware) can
+  become a challenge, if most of them are active concurrently. It
+  forces dbs to be set to `q=1`, migrating off `q!=1` requires
+  downtime, 10k bidirectional replications are going to need A LOT of
+  CPU and RAM. sharing documents among two or more users requires the
+  creation of yet more databases.
+
+Per-user document access aims to solve many of the above problems.
```
garrensmith: I think this paragraph and the goals are probably all thats needed for an abstract.

rfcs/010-per-document-access-control.md
```diff
@@ -0,0 +1,418 @@
...
+        _conflict handler.
+
+* document _ids are shared across all users. So only the first user who
+  creates the doc `_id: config` gets it. Applications need to ensure to
+  work around this and potentially prefix docs with the username before
+  writing/replicating them in.
+
+* _security members are allowed to write design docs, but they have to
+  have an `_access` field and those design docs with an `_access` field
+  are ignored on the server side. Db-admin ddocs get indexes built as
```
garrensmith: I think it might be clearer if you write this as: 
design doc's with  an `_access` field will be ignored in an access database

rfcs/010-per-document-access-control.md
```diff
@@ -0,0 +1,418 @@
...
+   * you can’t access their views, no view indexes are built, their
+     validate_doc_update functions do not run on db inserts.
+
+   * this allows full pouchdb / satellite db replication, but avoids
+     problems with having 10000s of VDUs or 10000s of view indexes.
+
+* users can not remove themselves from `_access`, nor can they remove
+  the `_access` property. They can only `DELETE` a doc.
+
+* If an existing doc changes the user mentioned in `_access` or an admin
```
garrensmith: What happens for a user that previously had access to the document? Is there a way to notify them that they have lost access?

rfcs/010-per-document-access-control.md
```diff
@@ -0,0 +1,418 @@
...
+* per-access views
+* differentiation between read and write access for documents
+* sharing individual documents between multiple users or groups.
+
+However, the design of this iteration aims to allow turning these
+non-goals into actual goals later.
+
+## Dramatis Personae
+
+*user*: a CouchDB-user, a record defined in the _users db identified by
```
garrensmith: Can you escape the `_users`. Everything below is ital

rfcs/010-per-document-access-control.md
```diff
@@ -0,0 +1,418 @@
...
+
+
+# Advantages and Disadvantages
+
+The downsides of this are the additional bookkeeping required in the
+newly created `by-access-seq` and `by-access-id` indexes. Given the
+resource requirements of the alternative db-per-user, this is a more
+than welcome trade-off.
+
+As a first iteration, this aims to tackle enough probelms to be useful
```
garrensmith: s/probelms/problems

garrensmith (changes requested):
Nice start to this.

rfcs/010-per-document-access-control.md
janl: thanks for your comments and I agree, this is mostly a paste from earlier, less structured, write-ups

rfcs/010-per-document-access-control.md
janl: yeah I must have missed this one, I have it that way further down

rfcs/010-per-document-access-control.md
janl: not as per this RFC, but we could consider a mode to `_changes`, e.g. `/_changes?access=true` that would include new rows with id/rev/revoked:true that clients could opt into.

garrensmith:
Sorry @janl I meant to add another note here the other day. Could you add a section on the data model for the new changes and _all_doc indexes.

mikerhodes (commented):
At the end of the abstract, you get to the goal of this RFC:

> * allow developers to build apps wihtout having to resort to using the
>   db-per-user pattern. Specifically PouchDB applications and CouchDB setups
>   with a central server/cluster and many independent satellite installations
>   with replication should be supported.

I think it's worth calling out at the very start instead that this is the goal
of this RFC because it's only possible to evaluate the design in light
of this goal.

I would then go on to specifically draw out that this document _doesn't_ want to
cover traditional three-tier style architectures, which really this
RFC leaves to be handled in the traditional way of, for example, using
views with keys prefixed by user name and having a trusted server.

I say this because when reading the RFC, for the first read through I was pretty
much like "this is crazy" before re-reading the abstract and starting to view
the RFC through the prism of supporting the hub/spoke type model, when suddenly
it made sense. Particularly the "don't build user views" thing was a total
confusion until I flipped my viewpoint. With the right context, it was easy to
see how these were a nice solution to a vexing problem.

The main worry that I have about this proposal is that it introduces a feature
that sounds like it's applicable to more scopes than it really is. We have
introduced a per-doc security model that only really works for one use-case.
While I can see this proposal doesn't _block_ figuring out how to make server
built views (reduces specifically) work with per-doc access, I fear that we end
up supporting a (valuable) mobile-first specific feature (safe replication from
a single database) using the terms of a much more generic feature (per document
access control) that we may never build. CouchApps are another feature that
suffered from being incomplete, in part because of our lack of server-side
per-doc controls :)

As noted in my comments, one thing I particularly wonder is whether these "spoke
only" design documents should be drawn out as a new API context rather than
inferring from `_access` whether they should be build on the server (this
behaviour seems to only make sense in the hub/spoke use-case). The behaviour
overloads "authn/z" concerns with server behaviours and I have found inferred
orthogonal behaviours usually cause a lot of pain later.

rfcs/010-per-document-access-control.md
```diff
@@ -0,0 +1,418 @@
...
+* admin: no changes
+* user: only the docs where `req.userCtx.name == _access: [$name]` plus
+ non-_access ddocs
+
+`/db/_sync_shards` TBD probably no changes
+
+`/db/_changes`
+
+* admin: no changes
+* user: only the docs where `req.userCtx.name == _access: [$name]` plus non-_access ddocs
```
nono: Is there a way to get the documents for which the access has been revoked from the changes feed?

Let's say that a user replicates the documents they have access in a pouchdb. Later, one of those documents is updated and the access for the user is revoked. If the user triggers a replication from CouchDB to PouchDB, they won't get this document from the changes feed (as I understand), and the document will still be in PouchDB with the old values. Is it correct?

wohali:
@janl Can we get this merged please? Thanks. I'm merging all of the RFCs that don't have any other sort of blockers; @garrensmith 's requests haven't been dismissed/changed so I'm not going to robo-merge this one.

janl:
closing this for now, I’ll open a new variant on the `couchdb` repo now that the docs repo is being retired.

PR #426, opened 2019-07-11
Clarify what compaction does

flimzy:


src/api/database/compact.rst
```diff
@@ -22,15 +22,15 @@
     Request compaction of the specified database. Compaction compresses the
     disk database file by performing the following operations:
 
-    - Writes a new, optimised, version of the database file, removing any
+    - Writes a new, optimized, version of the database file, removing any
```
wohali: I thought we were standardising on UK spellings?

flimzy: Are we? I missed that memo. But even so, I'm pretty sure that *-ize* is also acceptable British spelling.

flimzy: https://en.wikipedia.org/wiki/American_and_British_English_spelling_differences#-ise,_-ize_(-isation,_-ization)

flimzy: That said, I don't mind reverting that change if that's preferred.

wohali: It's fine, let it go

rnewson: -ize is not British spelling (we get our -ise from French).

src/api/database/compact.rst
```diff
@@ -22,15 +22,15 @@
...
       unused sections from the new version during write. Because a new file is
       temporarily created for this purpose, you may require up to twice the
       current storage space of the specified database in order for the
       compaction routine to complete.
-
-    - Removes old revisions of documents from the database, up to the
-      per-database limit specified by the ``_revs_limit`` database
-      parameter.
+    - Removes the bodies of any old, non-leaf revisions of documents from the
```
rnewson: "old" is redundant, "non-leaf" captures the meaning itself. We don't want to imply that wallclock time matters.

PR #431, opened 2019-08-13
feat: add example response for _bulk_get with two conflicts

janl:


src/api/database/bulk-api.rst
```diff
@@ -604,6 +604,58 @@ Sending multiple queries to a database
...
 
+    Example response with a conflicted document:
+
+    **Request**:
+
+    .. code-block:: http
+
+        POST /db/_bulk_get HTTP/1.1
+        Accept: application/json
+        Content-Type:application/json
```
flimzy: No space between header and value... minor issue at best :)

PR #441, opened 2019-09-24
initial reduce rfc

garrensmith:
## Overview

RFC for reduce indexes with FDB

## Checklist


rfcs/012-fdb-reduce.md
```diff
@@ -0,0 +1,190 @@
...
+            }
+        }
+    }
+
+```
+
+And it emits the following key/value results for the reduce function:
+
+```js
+    [2017, 03, 1] = 1
```
rnewson: what if more than one document emits the same key?

rfcs/012-fdb-reduce.md
```diff
@@ -0,0 +1,190 @@
...
+
+![figure 3: Traversing flow to return results for `group_level = 2`](images/SkExample3.png)
+*Figure 3:* Traversing flow to return results for `group_level = 2`
+
+A query with startkey/endkey would follow a similar process.  Start at the highest level traversing across until we exceed the startkey, then move down until we find the startkey or nearest node to it. Then follow the above process of traversing the skip list to return all results until we reach a node that is greater or equal to the endkey.
+
+#### Update
+
+To update the reduce index, we will use the same map id index that keeps track of what keys are associated with a document. When a document is deleted. The previous keys will be removed from level 0. If the reduce function is `_sum` or `_count`, an atomic update is then performed for all Node's above Level 0 that would have included the values for the keys deleted.
```
rnewson: "When a document is deleted." exists adrift from context, which sentence is it part of?

rfcs/012-fdb-reduce.md
```diff
@@ -0,0 +1,190 @@
...
...gure 2, with a reduce query of `group_level = 2`. We start at Level 4, we would traverse down to level 3, compare Node 0 and Node [2018, 03, 1]. They are not the same key for `group_level = 2` so we need to move to a lower level. We would perform the same action of comparing the current node with the next node to see if they are the same key until we find matching nodes or we get to level 0. In this case, we would reach level 0 and return Node [2017, 03, 1]. At Node [2017, 04, 1] it is possible to go back up to level 2 where we can compare [2017, 04, 1] and [2017, 04, 15]. Given they are the same, we can continue on level 1 and collect all the [2017, 04, x] keys. Once we have collected all of those. Those keys are run through a re-reduce before being returned. This process of comparing the current node with the next node would continue onwards. Each time comparing the current node with the next node. If they are the same, compare the Nodes at the next Level above until we reach a level where the keys are not equal. At that point, we would use the node values at the level just below that to move across, collecting all the equal keys to be run through a reducer before being returned.
```
kocolosk: Out of curiosity, why is the first Node `0` instead of the first key?

rfcs/012-fdb-reduce.md
```diff
@@ -0,0 +1,190 @@
...
+
+When updating a document, the initial delete process is followed to remove the existing keys that are no longer emitted for this document. The new keys are added at level 0. For level's above 0, the same distribution algorithm will be used to determine if the key/values are added to a level. If they are, then an aggregation of the nodes after this node but at the level below is performed to calculate the aggregation value stored for this node. The previous node's value is also recalculated. This continues up to the max level on the skip list. For `_sum` and `_count` this can be done as an atomic update. For other reduce functions this will be a re-reduce. If the new key/value is not added to the level, the value is aggregated with a node smaller than the key in that level.
+
+### Data model
+
+The data model for the skip list implementation will follow the same design as map indexes where we store one row for the key and one row for the Value.
```
kocolosk: I went back and read the map index RFC but didn't see any rationale for why we do this. It seems pretty wasteful. Is it because you want to reserve the full 100 KB for a possible emitted value?

kocolosk:
Good stuff @garrensmith . The upper levels of the skip list are going to be quite conflict-prone if there are multiple parallel updaters on the same database; do we need a dedicated discussion of how to mitigate that?

rfcs/012-fdb-reduce.md
```diff
@@ -0,0 +1,190 @@
...
+The simplest implementation is to perform the reduce function on the fly when a reduce query is requested.
+The map index will be read and the reduce aggregation performed on the keys as they are fetched from FoundationDB.
+This is by far the easiest to implement, but query performance will degrade as the database and index grows and could reach a point where the reduce query just stops working.
+
+### Option 2 - Precompute of all group-levels
+
+Another option is to precompute all group level's for a reduce function and store them as key/values in FoundationDB.
+This makes querying fast as the results are already calculated. The first difficulty comes when updating the index.
+For the built-in `_sum` and `_count` reduce functions, a single reduce calculation can be run and then applied to all group levels.
+For any custom reduce functions along with  `_stats` and `_approx_count_distinct`, the updating of each group level would be more complex as it requires reading all keys and running them through the reduce functions for all group levels before storing it back in FoundationDB.
```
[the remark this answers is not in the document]
garrensmith: Good point. I'll add that in so its more obvious

rfcs/012-fdb-reduce.md
```diff
@@ -0,0 +1,190 @@
...
+    [2018, 05, 1] = 1
+
+    [2019, 03, 1] = 1
+    [2018, 04, 1] = 1
+    [2018, 05, 1] = 1
+```
+
+#### Create
+
+To build the skip list, all keys will be added to level 0. Then each level up, a reduced number of keys will be added. For each level above 0, if a key/value is not added to that level, then that key's value is aggregated with the previous node in that row. Therefore each key/value node in a level is an aggregation of its key/value in level 0 and any key/values from the previous level that are greater than the node and smaller than the next node on that level.
```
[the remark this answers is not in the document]
garrensmith: Good idea. level = 0 is group=true

rfcs/012-fdb-reduce.md
```diff
@@ -0,0 +1,190 @@
...
+See figure 2 for an example of the listed keys added to a skip list.
+
+![figure 2: The emitted reduce keys added to a reduce skip list](images/SkExample2.png)
+*figure 2:* The emitted reduce keys added to a reduce skip lis
+
+The optimal number of levels and the correct distribution algorithm is unknown at this stage, these settings will be configurable so that we can test and determine the best fit.
+
+#### Query
+
+From the above figure 2, we can see that for a reduce query with `group = true` then level 0 will be used to return all the exact keys.
```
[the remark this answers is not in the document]
garrensmith: Yes, this did occur to me. I'm happy to maybe rename our levels alphabetically or anything else if that helps make this clearer

rfcs/012-fdb-reduce.md
garrensmith: I based that off of the rankedset implementation's I've seen. I'm definitely open to rather making it the first key. At this stage, I'm not sure which would be better.

rfcs/012-fdb-reduce.md
garrensmith: Good point. I need to update the map index to reflect this information. The reason we do this is once we encode the key so that the index sorts correctly. We cannot always decode it to exactly way the key that was emitted. Specifically for numbers which are encoded to doubles when stored in fdb but when returned to the user should often be integers. You had the discussion around number conversions on the mailing list. If we could convert all numbers to doubles I think we could remove the extra row.

garrensmith:
@kocolosk thanks for reviewing it. I agree, a separate discussion around parallel updates would be great. I wanted to tackle just one difficult task with this RFC. And I think the parallel updates discussion would be relevant for map indexes as well.

rfcs/012-fdb-reduce.md
kocolosk: Sorry, I should have been more specific. I know why we have to store the originally-emitted Key into an FDB Value. What I don't understand is why we dedicated a whole Value just to storing the emitted Key instead of packing it in with the emitted Value into a single FDB Value?

garrensmith: Ok right. Good point. Given that reduce values should be small. It does make a lot of sense to add the key and value in the same row. Thanks. I'll update the RFC.

kocolosk: Cool. For what it's worth the design you propose may not have the same "death spiral" effect with large reductions that CouchDB has today. The reason that happens today is because the btree chooses when to split based on the size of the inner node. Large reductions lead to a reduction of the branching factor and a deepening of the tree, which also causes more copies of the large reduction to be stored. And down we go :)

If you're careful in how you choose skip list levels we may end up much less sensitive to that effect.

rfcs/012-fdb-reduce.md
```diff
@@ -0,0 +1,190 @@
...
+
+![figure 3: Traversing flow to return results for `group_level = 2`](images/SkExample3.png)
+*Figure 3:* Traversing flow to return results for `group_level = 2`
+
+A query with startkey/endkey would follow a similar process.  Start at the highest level traversing across until we exceed the startkey, then move down until we find the startkey or nearest node to it. Then follow the above process of traversing the skip list to return all results until we reach a node that is greater or equal to the endkey.
```
alexmiller-apple: The algorithms to follow are described in a very CPU/Memory oriented way, and not an ordered key-value store oriented way.  Was that intentional for clarity?

Concretely, "traverse across until we exceed `startkey`" would just be "Read the key that is less than or equal to startkey at the highest level".  I think somewhere, it's at least worth mentioning that, while you are building a skiplist, you can randomly index into it unlike normal skiplists.

garrensmith: @alexmiller-apple I did intentionally focus the use of skiplists in a more theoretical concept. But you make a good point. I'll try and add some sudo code that uses fdb.

rfcs/012-fdb-reduce.md
garrensmith: The values would be rereduced together.

rfcs/012-fdb-reduce.md
garrensmith: Coming back to this. Having the first Node as `0` makes the starting point quite predictable and when we querying from the beginning of the index, one less key fetch to determine where to start.

garrensmith:
Thanks everyone for reviewing this. I've made fixes to the RFC based on that feedback.
I've also added a JS FDB Skiplist implementation. This is based off @alexmiller-apple comment that it would be better to have some actual code for this RFC. I would appreciate any feedback on that implementation as I'm going to use that work to drive the CouchDB implementation.

rfcs/012-fdb-reduce.md
garrensmith: Just adding another point on this because I was testing the implementation without having a starting point. Another issue we get here if we don't specify a starting every time we get a key that is the lowest in the index we would have to add it to all skiplevels. This means we would get a clump of keys at the beginning of the index.

rfcs/012-fdb-reduce.md
```diff
@@ -0,0 +1,1096 @@
...
+- `ViewId` id of a view defined in the design document
+- `SkipLevel` is the skip level the key/value is being stored for
+- `ReduceKey` is the encoded emitted keys
+- `RowType` indicates if the row is storing the emitted key or emitted value
+- `UnEncodedKey` the unencoded emitted keys
+- `Value` the reduce value for the emitted keys
+
+## FoundationDB Skip list implementation
+
+In Appendix A, is a full
```
nono: I think the end of the sentence is missing

PR #446, opened 2019-09-27
Fix typo

travispaul:
## Overview

Simple typo fix, remove duplicate of the word _"are"_.

### DIFF
--- a/src/best-practices/documents.rst
+++ b/src/best-practices/documents.rst
@@ -32,7 +32,7 @@ a few reasons:
   stored under duplicate ``_id``\ s. This could easily happen with intermediary
   proxies and cache systems that may not inform developers that the failed
   transaction is being retried.
-- ``_id``\ s are are the only unique enforced value within CouchDB so you might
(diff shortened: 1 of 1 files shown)


travispaul:
Also, I might be misunderstanding something but this doesn't sound right to me:

> If for any reason you miss the 200 OK reply from CouchDB, and storing the document is attempted again, you would end up with the same document content stored under duplicate _ids. This could easily happen with intermediary proxies and cache systems that may not inform developers that the failed transaction is being retried.

In the case of relying on CouchDB's autogenerated IDs, I would expect:

>  the same document content stored under **different** _ids.

My thought is that if a proxy retries a POST or PUT request, it will have the same document but a different generated ID.

wohali:
Hi @travispaul , thanks for your contribution to CouchDB! 🎆 🍾 

You're right, that should be `s/duplicate/multiple`. Would you like to add that to your PR? Then I can merge both changes!

travispaul:
Thanks for the fast response @wohali I pushed that change as well.

PR #470, opened 2020-01-08
Document rexi config settings

kocolosk:
## Overview

Document the config flags for the RPC library

## Related Pull Requests

apache/couchdb#2360 is a new config setting for 3.0

## Checklist

### DIFF
--- a/src/config/cluster.rst
+++ b/src/config/cluster.rst
@@ -85,3 +85,39 @@ Cluster Options
 
         [cluster]
         seedlist = couchdb@node1.example.com,couchdb@node2.example.com
+
+RPC Performance Tuning
+======================
+
+.. config:section:: rexi :: Internal RPC Tuning
+
+    CouchDB uses distributed Erlang to communicate between nodes in a cluster.
+    The ``rexi`` library provides an optimized RPC mechanism over this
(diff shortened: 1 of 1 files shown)


src/config/cluster.rst
```diff
@@ -85,3 +85,39 @@ Cluster Options
...
+    The local RPC server will buffer messages if a remote node goes unavailable.
+    This flag determines how many messages will be buffered before the local
+    server starts dropping messages. Default value is ``2000``.
+
+    .. config:option:: server_per_node
+
+    Early versions of this RPC library used a single local Erlang process for
+    sending messages to all other nodes in a cluster. The current default is to
+    spawn a local gen_server process for each node in the cluster. It is not
+    recommended to change this setting from the default.
```
flimzy: Is this history lesson helpful to CouchDB users? Could we simplify this paragraph with something like:

> By default, ``rexi`` will spawn one local ``gen_server`` process for each node in the cluster. This setting changes the number of ``gen_server`` processes per node. It is not recommended to change this setting from the default.

kocolosk: Good call. I'm sitting here telling the user why she should ignore this option, and wasting even more of her time in doing so! I took another pass with essentially your recommendation.

PR #482, opened 2020-01-22
Document enable_database_recovery

### LINKED ISSUE #284: Configuration option `[couchdb] enable_database_recovery` is not documented
The configuration option `[couchdb] enable_database_recovery` is not documented. It is used in CouchDB [here](https://github.com/apache/couchdb/blob/f999071ec6b347b27526047cc16a00811e21fdd9/src/couch/src/couch_file.erl#L236-L248).

The configuration option, when set to true, causes database deletion to rename the database file rather than actually delete it, allowing an admin to restore the database from the backup file.

willholley:
## Overview

Adds documentation for couchdb.enable_database_recovery. 

## GitHub issue number

Fixes #284 

## Related Pull Requests

Description lifted from https://github.com/apache/couchdb/pull/1283

## Checklist


src/config/couchdb.rst
```diff
@@ -194,3 +194,16 @@ Base CouchDB Options
...
            it is advisable to review those settings and adjust them accordingly.
+
+    .. config:option:: enable_database_recovery :: Enable database recovery
+
+        Enable this to only "soft-delete" databases when
+        :ref:`DELETE /{db} <api/db>` DELETE  requests are made. This will place
+        a ``.recovery`` directory in your data directory and move deleted
+        databases/shards there instead. You can then manually delete these
+        files later, as desired.
+
```
jiangphcn: The same for full text index, i.e.  index generated by dreyfys/clouseau. The directory containing full text index will be renamed from `foo.1234567890` to `foo.20170912.092828.deleted.1234567890` for recovery.

willholley: @jiangphcn just to clarify does `enable_database_recovery` toggles index soft deletion as well or are they configured independently? Do views follow the same behavior?

flimzy: What happens in the case of deleting multiple databases with the same name?  i.e. create "foo", soft-delete "foo", re-create "foo", soft-delete "foo".

Is the original "foo" lost, or are these disambiguated in some way?

jiangphcn: @willholley one setting for all, at least for full text index https://github.com/apache/couchdb/blob/0155bd884bd8c455cc4057811cf67035b04c427f/src/dreyfus/src/dreyfus_index_manager.erl#L122. Views follow the same behavior.

willholley: > What happens in the case of deleting multiple databases with the same name? i.e. create "foo", soft-delete "foo", re-create "foo", soft-delete "foo".
good question and I don't know - I just copied the description from default.ini

jiangphcn: there will be timestamp to distinguish them with the same database name, such as `foo.20170912.092828.deleted.1234567890`.

PR #483, opened 2020-01-22
Fix typo in `_scheduler/docs`

bessbd:
## Overview

The response description for `_scheduler/docs` says there's a key `last_update`, however there isn't one. There is a `last_updated` instead, just as it shows in the example right below.

This change fixes the issue described above.

## Testing recommendations

Query `_scheduler/docs` and observe the response.

## Checklist


bessbd:
Thank you for the review and the merge, @iilyak

PR #499, opened 2020-02-27
Replace foo with sensor-260

fox1t:
## Overview
Little typo on the example.


wohali:
@fox1t Thank you for your contribution to Apache CouchDB! :clinking_glasses: :+1: :100:

PR #508, opened 2020-03-14
Fixes paths in sections titles and example snippet

fox1t:
## Checklist


wohali:
@fox1t Thanks!

PR #510, opened 2020-03-20
Attempt to explain partitioning better.

xrobau:
## Overview

Documentation and wording around partitioning is potentially confusing

## Checklist

Signed-Off-By: Rob Thomas <xrobau@linux.com>

### DIFF
--- a/src/partitioned-dbs/index.rst
+++ b/src/partitioned-dbs/index.rst
@@ -105,14 +105,20 @@ sensor readings from an IoT field monitoring service. In this particular
 use case, it's quite logical to group all documents by their ``sensor_id``
 field. In this case, we would call the ``sensor_id`` the partition.
 
+A partition offers two benefits. Firstly, it allows you to simplify the
+workload on the query engine, by allowing the engine to only query shards
+that it needs to query. Secondly, it allows the database to scale in more
+predictable ways.
+
 A good partition has two basic properties. First, it should have a high
 cardinality. That is, a large partitioned database should have many more
-partitions than documents in any single partition. A database that has
-a single partition would be an anti-pattern for this feature. Secondly,
-the amount of data per partition should be "small". The general
-recommendation is to limit individual partitions to less than ten
-gigabytes (10 GB) of data. Which, for the example of sensor documents,
-equates to roughly 60,000 years of data.
+partitions than documents, compared to an unpartitioned database. A database
+that has a single partition would be an anti-pattern for this feature, as
+one of the major bonuses of partitioning is that it allows index lookups to
+iterate over a smaller B-Tree. Secondly, the amount of data per partition
+should be "small". The general recommendation is to limit individual
+partitions to less than ten gigabytes (10 GB) of data. Which, for the example
+of sensor documents, equates to roughly 60,000 years of data.
 
 Why use partitions?
 ===================
@@ -129,6 +135,34 @@ given shard range when executing a query instead of executing
 the query across all ``q`` shards in the database. This mean that you do
 not have to wait for all ``q`` shards to respond, which is both
 efficient and faster.
+
+In other words, you can use partitions to help pre-process views. As another
+example, you may be recording traffic to various websites, and 95% of your
+queries are based around 'what happened in relation to website example.com'.
+By partitioning your database with the key ``example.com:uuid``, and using
+the ``/dbname/_partition/example.com`` query, you are significantly simplifying
+the work the query engine has to do.
+
+When to NOT use partitions?
+===========================
+
+If you can not split your data using a common partition, or you end up with
+a small number of partitions, it is more efficient to use seperate databases.
+Following on with the example above, if you are only monitoring 10 websites,
+it is far more efficient to have 10 databases, rather than 1 database with
+10 partitions, with 10,000,000 entries in one partition, 500 in another, and
+10,000 in all the others.
+
+As partitioning relates to which shard the data is placed into, you want to
+roughly balance the amount of data in each partition, but you should think
+of that as a design guideline when you are thinking about what to partition
+your data on.
+
+Additinally, if you are not using the ``/dbname/_partition/...`` endpoint
+for a majority of your queries, you are also potentially losing performance,
+because you won't get the automatic load-balancing of documents over shards
+that a non partitioned database gives you.
+
 
 Partitions By Example
 =====================
--- a/src/partitioned-dbs/index.rst
+++ b/src/partitioned-dbs/index.rst
@@ -163,7 +163,6 @@ for a majority of your queries, you are also potentially losing performance,
 because you won't get the automatic load-balancing of documents over shards
 that a non partitioned database gives you.
 
-
 Partitions By Example
 =====================
 


src/partitioned-dbs/index.rst
```diff
@@ -105,14 +105,20 @@ sensor readings from an IoT field monitoring service. In this particular
...
+
 A good partition has two basic properties. First, it should have a high
 cardinality. That is, a large partitioned database should have many more
-partitions than documents in any single partition. A database that has
-a single partition would be an anti-pattern for this feature. Secondly,
-the amount of data per partition should be "small". The general
-recommendation is to limit individual partitions to less than ten
-gigabytes (10 GB) of data. Which, for the example of sensor documents,
-equates to roughly 60,000 years of data.
+partitions than documents, compared to an unpartitioned database. A database
```
flimzy: This sentence is confusing.  More partitions than documents?

src/partitioned-dbs/index.rst
```diff
@@ -130,6 +136,33 @@ the query across all ``q`` shards in the database. This mean that you do
...
 efficient and faster.
 
+In other words, you can use partitions to help pre-process views. As another
+example, you may be recording traffic to various websites, and 95% of your
+queries are based around 'what happened in relation to website example.com'.
+By partitioning your database with the key ``example.com:uuid``, and using
+the ``/dbname/_partition/example.com`` query, you are significantly simplifying
+the work the query engine has to do.
+
+When to NOT use partitions?
```
flimzy:
```suggestion
When to NOT use partitions
```

src/partitioned-dbs/index.rst
```diff
@@ -130,6 +136,33 @@ the query across all ``q`` shards in the database. This mean that you do
...
+===========================
+
+If you can not split your data using a common partition, or you end up with
```
flimzy:
```suggestion
If you cannot split your data using a common partition, or you end up with
```

src/partitioned-dbs/index.rst
```diff
@@ -105,14 +105,20 @@ sensor readings from an IoT field monitoring service. In this particular
 use case, it's quite logical to group all documents by their ``sensor_id``
 field. In this case, we would call the ``sensor_id`` the partition.
 
+A partition offers two benefits. Firstly, it allows you to simplify the
+workload on the query engine, by allowing the engine to only query shards
```
wohali: I'm not sure "simplify the workload" is clearer here. What it's doing is allowing the user to group data in a way that is understood _a priori_, a.k.a. ahead of time. Since most searches (theoretically) are being done only within one given partition, queries are _optimised_ by only having to run within any one given partition at a time.

src/partitioned-dbs/index.rst
```diff
@@ -105,14 +105,20 @@ sensor readings from an IoT field monitoring service. In this particular
...
-partitions than documents in any single partition. A database that has
-a single partition would be an anti-pattern for this feature. Secondly,
-the amount of data per partition should be "small". The general
-recommendation is to limit individual partitions to less than ten
-gigabytes (10 GB) of data. Which, for the example of sensor documents,
-equates to roughly 60,000 years of data.
+partitions than documents, compared to an unpartitioned database. A database
+that has a single partition would be an anti-pattern for this feature, as
+one of the major bonuses of partitioning is that it allows index lookups to
+iterate over a smaller B-Tree. Secondly, the amount of data per partition
```
wohali: Would prefer the discussion didn't get into the implementation detail that, in CouchDB 3.x, storage is B-Tree based. This won't be the case with FoundationDB (CouchDB 4.0). It'd be sufficient to say that "it allows index lookups to occur over a small set of documents."

src/partitioned-dbs/index.rst
wohali: Hm, you lose the subtlety here of "significantly more partitions than documents in a single partition." I don't think that's necessarily a good thing.

1,000,000 documents split into 10,000 partitions is 100 docs/partition. That would satisfy the recommendation in the original text, but not your new definition, since 10,000 partitions isn't many more partitions than documents in the original 1,000,000 document database.

src/partitioned-dbs/index.rst
```diff
@@ -130,6 +136,33 @@ the query across all ``q`` shards in the database. This mean that you do
 not have to wait for all ``q`` shards to respond, which is both
 efficient and faster.
 
+In other words, you can use partitions to help pre-process views. As another
+example, you may be recording traffic to various websites, and 95% of your
+queries are based around 'what happened in relation to website example.com'.
+By partitioning your database with the key ``example.com:uuid``, and using
+the ``/dbname/_partition/example.com`` query, you are significantly simplifying
+the work the query engine has to do.
```
wohali: Compare and contrast this argument with that we make for Mango partial indexes - where we artificially limit the documents in a Mango index based on an initial condition. How does this differ from our suggestion there? The text is unclear to the prospective user.

src/partitioned-dbs/index.rst
wohali: Not exactly good grammar. Why not: "Why not use partitions?" This would be parallel structure with the previous heading.

src/partitioned-dbs/index.rst
wohali: `cannot` is one word.

src/partitioned-dbs/index.rst
```diff
@@ -130,6 +136,33 @@ the query across all ``q`` shards in the database. This mean that you do
...
+
+When to NOT use partitions?
+===========================
+
+If you can not split your data using a common partition, or you end up with
+a small number of partitions, it is more efficient to use seperate databases.
```
wohali: `separate`

src/partitioned-dbs/index.rst
wohali: What is a "common partition?"

src/partitioned-dbs/index.rst
```diff
@@ -130,6 +136,33 @@ the query across all ``q`` shards in the database. This mean that you do
...
+Following on with the example above, if you are only monitoring 10 websites,
```
wohali: It is confusing that with your PR, we now have 2 examples in this documentation section: website monitoring, and sensor monitoring. Are you able to unify your example with the sensor monitoring example?

Or, if not, can you clarify what you're referring to with "the example above" in this line?

src/partitioned-dbs/index.rst
```diff
@@ -130,6 +136,33 @@ the query across all ``q`` shards in the database. This mean that you do
...
+it is far more efficient to have 10 databases, rather than 1 database with
+10 partitions, with 10,000,000 entries in one partition, 500 in another, and
+10,000 in all the others.
```
wohali: It might be more straightforward to talk about "consistently sized partitions" .

src/partitioned-dbs/index.rst
```diff
@@ -130,6 +136,33 @@ the query across all ``q`` shards in the database. This mean that you do
...
+
+As partitioning relates to which shard the data is placed into, you want to
+roughly balance the amount of data in each partition, but you should think
+of that as a design guideline when you are thinking about what to partition
+your data on.
```
wohali: This is an extremely awkward, run-on sentence. I like the idea you're introducing here, but please try to rephrase it.

src/partitioned-dbs/index.rst
```diff
@@ -130,6 +136,33 @@ the query across all ``q`` shards in the database. This mean that you do
...
+
+Additinally, if you are not using the ``/dbname/_partition/...`` endpoint
+for a majority of your queries, you are also potentially losing performance,
+because you won't get the automatic load-balancing of documents over shards
+that a non partitioned database gives you.
```
wohali: Same point as the previous paragraph; besides the misspelling, this is awkwardly phrased. I like the idea of "Oh, you're not using the `_partition` specific endpoints most of the time? You probably don't need a partitioned DB."

src/partitioned-dbs/index.rst
```diff
@@ -105,14 +105,20 @@ sensor readings from an IoT field monitoring service. In this particular
 use case, it's quite logical to group all documents by their ``sensor_id``
 field. In this case, we would call the ``sensor_id`` the partition.
 
+A partition offers two benefits. Firstly, it allows you to simplify the
+workload on the query engine, by allowing the engine to only query shards
+that it needs to query. Secondly, it allows the database to scale in more
+predictable ways.
+
```
flimzy: Several thoughts on this paragraph:

1. It's not technically accurate to say "A partition", since the whole point is multiple partitions.
2. I think there can be more than _only_ two benefits, so we shouldn't imply that these are the only benefits.
3. It also doesn't really "simplify" the workload, but rather reduces it (partitioning is inherently more complex, actually, at least in the mathematical sense)
4. "only query the shards it needs to query" doesn't seem clear to me. It sounds like a tautology.

I would suggest something like:

> Partitioning offers two primary benefits. First, it reduces the workload on the query engine by
> allowing it to query only a subset of shards. Second, it allows the database to scale in more 
> predictable ways.

src/partitioned-dbs/index.rst
flimzy: To be consistent with "First,", earlier in the paragraph, we should also say "Second", not "Secondly". ("Firstly" and "Secondly" are awkward in this context anyway, as they're adverbs, not adjectives, which subtly changes the meaning of the sentence, but I digress... :smile: ).

```suggestion
iterate over a smaller B-Tree. Second, the amount of data per partition
```

src/partitioned-dbs/index.rst
```diff
@@ -105,14 +105,20 @@ sensor readings from an IoT field monitoring service. In this particular
...
-the amount of data per partition should be "small". The general
-recommendation is to limit individual partitions to less than ten
-gigabytes (10 GB) of data. Which, for the example of sensor documents,
-equates to roughly 60,000 years of data.
+partitions than documents, compared to an unpartitioned database. A database
+that has a single partition would be an anti-pattern for this feature, as
+one of the major bonuses of partitioning is that it allows index lookups to
+iterate over a smaller B-Tree. Secondly, the amount of data per partition
+should be "small". The general recommendation is to limit individual
+partitions to less than ten gigabytes (10 GB) of data. Which, for the example
```
flimzy: What is this recommendation based on?

Would it not make more sense to make a recommendation based on index size than on data size? Given that index size isn't trivial to determine, maybe basing this recommendation on number of documents (which is usually a closer proxy to index size) would make sense?

@wohali probably has better insight into the technical details here than me.

src/partitioned-dbs/index.rst
flimzy: Possible suggestion:

> To balance the data evenly across shards, it is useful to consider the partition key when designing your database.

src/partitioned-dbs/index.rst
flimzy: Possible suggestion:

> If you find that you are rarely using ``/dbname/_partition/...`` endpoints, the use of partitions may actually hurt performance. This is because CouchDB is unable to automatically distribute documents and associated queries across shards in the most efficient way.

src/partitioned-dbs/index.rst
wohali: https://blog.cloudant.com/2019/03/05/Partition-Databases-Introduction.html has an excellent introduction, see under _What makes a good partition key?_

We can't cut and paste that text, but it should help.

Note it's a 4-part blog series; it'd be good to read all of it before moving ahead.

wohali:
@xrobau Thanks again for your help on this. Unfortunately, we can't accept this PR in this state.

I'm going to close the PR for now. If you can find the time to incorporate the suggestions outlined above, we're happy to work through another PR with you - just let us know.

Once again thank you for your contribution to Apache CouchDB!

PR #518, opened 2020-04-08
Fix _all_docs/queries example result

bessbd:
## Overview

The example result for _all_docs/queries did not match the example request and it was incorrect in a sense that the values `id` fields were not matching the respective values of `key` fields and there was a document returned that was not specified in the query.

This change fixes the issue described above by editing the ids so that they match the keys now and removing the element that was not specified in the request.

## Checklist


bessbd:
Thank you for the review and the merge, @flimzy and @nickva !

PR #520, opened 2020-04-10
Add note of ddoc autoupdate:false to ken config

Terreii:
## Overview

When looking for ways to control background indexing, users usually land on [3.8. Background Indexing](https://docs.couchdb.org/en/3.0.0/config/indexbuilds.html).
But if they are looking for a way to disable background indexing of a specific ddoc or view, that option (`autoupdate`) is documented in the [ddoc API](https://docs.couchdb.org/en/3.0.0/api/ddoc/common.html#put--db-_design-ddoc).

This pull request adds a note about `autoupdate:false` and a link from background indexing to the ddoc API.

## Testing recommendations

Decide if a `Note` or `See also` is more appropriate.

## GitHub issue number

There is no bug. It is probably not a significant change.

## Checklist

### DIFF
--- a/src/config/indexbuilds.rst
+++ b/src/config/indexbuilds.rst
@@ -59,3 +59,7 @@ over specific database shard files. The key must be the exact name of the shard 
 
         [ken.ignore]
         shards/00000000-1fffffff/mydb.1567719095 = true
+
+    .. note::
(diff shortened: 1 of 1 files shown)


Terreii:
Hi all, this Pull request is in reaction to a question that did happen in the CouchDB Slack.

PR #524, opened 2020-04-19
JWT documentation

rnewson:
## Overview

Documentation for the optional JWT authentication handler

## Testing recommendations

N/A.

## GitHub issue number

N/A

## Related Pull Requests

https://github.com/apache/couchdb/pull/2648

## Checklist

- [x] Update [rebar.config.script](https://github.com/apache/couchdb/blob/master/rebar.config.script) with the commit hash once this PR is rebased and merged

### DIFF
--- a/src/api/server/authn.rst
+++ b/src/api/server/authn.rst
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
 
 Note that you don't need to request :ref:`session <api/auth/session>`
 to be authenticated by this method if all required HTTP headers are provided.
+
+.. _api/auth/jwt:
+
+JWT Authentication
+====================
+
+.. note::
+    To use this authentication method make sure that the
+    ``{chttpd_auth, jwt_authentication_handler}`` value in added to the
+    list of the active :config:option:`chttpd/authentication_handlers`:
+
+    .. code-block:: ini
+
+        [chttpd]
+        authentication_handlers = {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, jwt_authentication_handler}, {chttpd_auth, default_authentication_handler}
+
+`JWT authentication` enables CouchDB to use externally generated JWT tokens
+instead of defining users or roles in the _users database.
+
+The JWT authentication handler requires that all JWT tokens are signed by a key that
+CouchDB has been configured to trust (there is no support for JWT's "NONE" algorithm).
+
+Additionally, CouchDB can be configured to reject JWT tokens that are missing a
+configurable set of claims (e.g, a CouchDB administrator could insist on the exp claim).
+
+All claims presented in a JWT token are validated if presented, regardless of whether they
+are required.
+
+Two new sections of config have been introduced to configure JWT authentication;
+
+.. code-block:: ini
+
+    ;[jwt_auth]
+    ; List of claims to validate
+    ; required_claims =
+
+The `required_claims` config setting is a comma-separate list of additional mandatory
+JWT claims that much be present in any presented JWT token. A 400 Bad Request is sent
+if any are missing.
+
+The `alg` claim is mandatory as it used to lookup the correct key for verifying the
+signature.
+
+The `sub` claim is mandatory and is used as the CouchDB user's name if the JWT token
+is valid.
+
+A private claim called `_couchdb.roles` is optional. If presented, it is a JSON
+array of strings, as is used as the CouchDB user's roles list if the JWT token
+is valid.
+
+.. code-block:: ini
+
+    ; [jwt_keys]
+    ; Configure at least one key here if using the JWT auth handler.
+    ; If your JWT tokens do not include a "kid" attribute, use "_default"
+    ; as the config key, otherwise use the kid as the config key.
+    ; Examples
+    ; hmac:_default = aGVsbG8=
+    ; hmac:foo = aGVsbG8=
+    ; The config values can represent symmetric and asymmetrics keys.
+    ; For symmetrics keys, the value is base64 encoded;
+    ; hmac:_default = aGVsbG8= # base64-encoded form of "hello"
+    ; For asymmetric keys, the value is the PEM encoding of the public
+    ; key with newlines replaced with the escape sequence \n.
+    ; rsa:foo = -----BEGIN PUBLIC KEY-----\nMIIBIjAN...IDAQAB\n-----END PUBLIC KEY-----\n
+    ; ec:bar = -----BEGIN PUBLIC KEY-----\nMHYwEAYHK...AzztRs\n-----END PUBLIC KEY-----\n
+
+The `jwt_key` section lists all the keys that this CouchDB server trusts. You
+should ensure that all nodes of your cluster has the same list.
+
+JWT tokens that do not include a `kid` claim will be validated against the
+`$alg:_default` key.
+
+It is mandatory to specify the algorithm associated with every key for security
+reasons (notably presenting a HMAC signed token using a RSA or EC public key
+that the server trusts:
+https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/).
+
+**Request**:
+
+.. code-block:: http
+
+    GET /_session HTTP/1.1
+    Host: localhost:5984
+    Accept: application/json
+    Content-Type: application/json; charset=utf-8
+    Authorization: Bearer <JWT token>
+
+**Response**:
+
+.. code-block:: http
+
+    HTTP/1.1 200 OK
+    Cache-Control: must-revalidate
+    Content-Length: 188
+    Content-Type: application/json
+    Date: Sun, 19 Apir 2020 08:29:15 GMT
+    Server: CouchDB (Erlang/OTP)
+
+    {
+        "info": {
+            "authenticated": "jwt",
+            "authentication_db": "_users",
+            "authentication_handlers": [
+                "cookie",
+                "proxy",
+                "default"
+            ]
+        },
+        "ok": true,
+        "userCtx": {
+            "name": "foo",
+            "roles": [
+                "users",
+                "blogger"
+            ]
+        }
+    }
+
+Note that you don't need to request :ref:`session <api/auth/session>`
+to be authenticated by this method if the required HTTP header is provided.
--- a/src/api/server/authn.rst
+++ b/src/api/server/authn.rst
@@ -376,12 +376,6 @@ are required.
 
 Two sections of config exist to configure JWT authentication;
 
-.. code-block:: ini
-
-    ;[jwt_auth]
-    ; List of claims to validate
-    ; required_claims =
-
 The :config:option:`required_claims <jwt_auth/required_claims>` config
 setting is a comma-separated list of additional mandatory JWT claims
 that must be present in any presented JWT token. A `:code 400:Bad
--- a/src/api/server/authn.rst
+++ b/src/api/server/authn.rst
@@ -381,9 +381,10 @@ Two sections of config exist to configure JWT authentication;
     ; List of claims to validate
     ; required_claims =
 
-The `required_claims` config setting is a comma-separated list of additional mandatory
-JWT claims that must be present in any presented JWT token. A `:code 400:Bad Request` is sent
-if any are missing.
+The :config:option:`required_claims <jwt_auth/required_claims>` config
+setting is a comma-separated list of additional mandatory JWT claims
+that must be present in any presented JWT token. A `:code 400:Bad
+Request` is sent if any are missing.
 
 The ``alg`` claim is mandatory as it used to lookup the correct key for verifying the
 signature.
@@ -391,9 +392,9 @@ signature.
 The ``sub`` claim is mandatory and is used as the CouchDB user's name if the JWT token
 is valid.
 
-A private claim called ``_couchdb.roles`` is optional. If presented, it is a JSON
-array of strings, as is used as the CouchDB user's roles list if the JWT token
-is valid.
+A private claim called ``_couchdb.roles`` is optional. If presented,
+it is a JSON array of strings, it is used as the CouchDB user's roles
+list as long as the JWT token is valid.
 
 .. code-block:: ini
 
--- a/src/api/server/authn.rst
+++ b/src/api/server/authn.rst
@@ -393,7 +393,7 @@ The ``sub`` claim is mandatory and is used as the CouchDB user's name if the JWT
 is valid.
 
 A private claim called ``_couchdb.roles`` is optional. If presented,
-it is a JSON array of strings, it is used as the CouchDB user's roles
+as a JSON array of strings, it is used as the CouchDB user's roles
 list as long as the JWT token is valid.
 
 .. code-block:: ini
--- a/src/api/server/authn.rst
+++ b/src/api/server/authn.rst
@@ -368,7 +368,8 @@ The JWT authentication handler requires that all JWT tokens are signed by a key 
 CouchDB has been configured to trust (there is no support for JWT's "NONE" algorithm).
 
 Additionally, CouchDB can be configured to reject JWT tokens that are missing a
-configurable set of claims (e.g, a CouchDB administrator could insist on the ``exp`` claim).
+configurable set of claims (e.g, a CouchDB administrator could insist on the ``exp``
+claim).
 
 All claims presented in a JWT token are validated if presented, regardless of whether they
 are required.
--- a/src/api/server/authn.rst
+++ b/src/api/server/authn.rst
@@ -352,7 +352,7 @@ JWT Authentication
 ====================
 
 .. note::
-    To use this authentication method make sure that the
+    To use this authentication method, make sure that the
     ``{chttpd_auth, jwt_authentication_handler}`` value is added to the
     list of the active :config:option:`chttpd/authentication_handlers`:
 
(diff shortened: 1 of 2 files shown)


src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+    To use this authentication method make sure that the
+    ``{chttpd_auth, jwt_authentication_handler}`` value in added to the
+    list of the active :config:option:`chttpd/authentication_handlers`:
+
+    .. code-block:: ini
+
+        [chttpd]
+        authentication_handlers = {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, jwt_authentication_handler}, {chttpd_auth, default_authentication_handler}
+
+`JWT authentication` enables CouchDB to use externally generated JWT tokens
```
flimzy, line 364:
```suggestion
`JWT authentication` enables CouchDB to use externally-generated JWT tokens
```

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+instead of defining users or roles in the _users database.
```
flimzy, line 365:
```suggestion
instead of defining users or roles in the ``_users`` database.
```

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+
+The JWT authentication handler requires that all JWT tokens are signed by a key that
+CouchDB has been configured to trust (there is no support for JWT's "NONE" algorithm).
+
+Additionally, CouchDB can be configured to reject JWT tokens that are missing a
+configurable set of claims (e.g, a CouchDB administrator could insist on the exp claim).
```
flimzy, line 371: Not sure if this formatting is appropriate, since I'm not overly familiar with JWT

```suggestion
configurable set of claims (e.g, a CouchDB administrator could insist on the ``exp`` claim).
```

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+
+All claims presented in a JWT token are validated if presented, regardless of whether they
+are required.
+
+Two new sections of config have been introduced to configure JWT authentication;
```
flimzy, line 376: [Optional] More future-proof wording?

```suggestion
Two sections of config exist to configure JWT authentication;
```

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+
+.. code-block:: ini
+
+    ;[jwt_auth]
+    ; List of claims to validate
+    ; required_claims =
+
+The `required_claims` config setting is a comma-separate list of additional mandatory
```
flimzy, line 384:
```suggestion
The `required_claims` config setting is a comma-separated list of additional mandatory
```

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+JWT claims that much be present in any presented JWT token. A 400 Bad Request is sent
```
flimzy, line 385:
```suggestion
JWT claims that must be present in any presented JWT token. A ``400 Bad Request`` is sent
```

Or consider linking to https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.1, as some other references to HTTP status codes do.

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+if any are missing.
+
+The `alg` claim is mandatory as it used to lookup the correct key for verifying the
```
flimzy, line 388: Maybe?  Should probably be consistent with "exp" on line 371 above, whichever formatting we choose.
```suggestion
The ``alg`` claim is mandatory as it used to lookup the correct key for verifying the
```

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+    ; For asymmetric keys, the value is the PEM encoding of the public
+    ; key with newlines replaced with the escape sequence \n.
+    ; rsa:foo = -----BEGIN PUBLIC KEY-----\nMIIBIjAN...IDAQAB\n-----END PUBLIC KEY-----\n
+    ; ec:bar = -----BEGIN PUBLIC KEY-----\nMHYwEAYHK...AzztRs\n-----END PUBLIC KEY-----\n
+
+The `jwt_key` section lists all the keys that this CouchDB server trusts. You
+should ensure that all nodes of your cluster has the same list.
+
+JWT tokens that do not include a `kid` claim will be validated against the
+`$alg:_default` key.
```
flimzy, line 419: Is `$alg` intentional here? (Again, I know nothing about JWTs).  If so, does it belong in the other places `_default` (or other keys) are mentioned?

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+    ; The config values can represent symmetric and asymmetrics keys.
+    ; For symmetrics keys, the value is base64 encoded;
+    ; hmac:_default = aGVsbG8= # base64-encoded form of "hello"
+    ; For asymmetric keys, the value is the PEM encoding of the public
+    ; key with newlines replaced with the escape sequence \n.
+    ; rsa:foo = -----BEGIN PUBLIC KEY-----\nMIIBIjAN...IDAQAB\n-----END PUBLIC KEY-----\n
+    ; ec:bar = -----BEGIN PUBLIC KEY-----\nMHYwEAYHK...AzztRs\n-----END PUBLIC KEY-----\n
+
+The `jwt_key` section lists all the keys that this CouchDB server trusts. You
+should ensure that all nodes of your cluster has the same list.
```
flimzy, line 416:
```suggestion
should ensure that all nodes of your cluster have the same list.
```

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+
+JWT tokens that do not include a `kid` claim will be validated against the
+`$alg:_default` key.
+
+It is mandatory to specify the algorithm associated with every key for security
+reasons (notably presenting a HMAC signed token using a RSA or EC public key
```
flimzy, line 422:
```suggestion
reasons (notably presenting a HMAC-signed token using an RSA or EC public key
```

wohali:
@rnewson consistent bug that needs fixing: .rst uses double backticks to escape preformatted/monospace content, not single backticks.

src/api/server/authn.rst
rnewson: I don't think this should be hyphenated.

src/api/server/authn.rst
rnewson: it is intentional but not as clear as I'd like. the point is that the algorithm (HMAC vs RSA, etc) is part of the identity of the secret.

wohali: How about appending `(e.g. ``hmac_default`` or ``rsa_default``)`? I might have the syntax wrong, don't know if the colon is part of it or not.

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+    To use this authentication method make sure that the
+    ``{chttpd_auth, jwt_authentication_handler}`` value is added to the
+    list of the active :config:option:`chttpd/authentication_handlers`:
+
+    .. code-block:: ini
+
+        [chttpd]
+        authentication_handlers = {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, jwt_authentication_handler}, {chttpd_auth, default_authentication_handler}
+
+`JWT authentication` enables CouchDB to use externally generated JWT tokens
```
wohali, line 364: `JWT authentication` needs different formatting, either double-backticks, or perhaps you meant bold/otherwise?

If you meant it to be a [phrase reference](https://docutils.sourceforge.io/docs/user/rst/quickref.html#hyperlink-targets) then it's incomplete.

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+
+Two sections of config exist to configure JWT authentication;
+
+.. code-block:: ini
+
+    ;[jwt_auth]
+    ; List of claims to validate
+    ; required_claims =
+
+The `required_claims` config setting is a comma-separated list of additional mandatory
```
wohali, line 384: Please use the correct syntax here, for instance:

```
:config:option:`X-Auth-CouchDB-Roles <couch_httpd_auth/x_auth_roles>`
```

This gives you the correct formatting, and also hyperlinks to the config portion of the docs.

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+JWT claims that must be present in any presented JWT token. A ``400 Bad Request`` is sent
```
wohali, line 385: Please use the `:code 400: Invalid request` syntax so this is correctly formatted and hyperlinked. (or is it really Bad Request? everywhere else in the docs we use Invalid Request.)

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+if any are missing.
+
+The ``alg`` claim is mandatory as it used to lookup the correct key for verifying the
+signature.
+
+The `sub` claim is mandatory and is used as the CouchDB user's name if the JWT token
```
wohali, line 391: Double quotation marks please.

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+is valid.
+
+A private claim called `_couchdb.roles` is optional. If presented, it is a JSON
```
wohali, line 394: Double quotation marks

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+    ; hmac:foo = aGVsbG8=
+    ; The config values can represent symmetric and asymmetrics keys.
+    ; For symmetrics keys, the value is base64 encoded;
+    ; hmac:_default = aGVsbG8= # base64-encoded form of "hello"
+    ; For asymmetric keys, the value is the PEM encoding of the public
+    ; key with newlines replaced with the escape sequence \n.
+    ; rsa:foo = -----BEGIN PUBLIC KEY-----\nMIIBIjAN...IDAQAB\n-----END PUBLIC KEY-----\n
+    ; ec:bar = -----BEGIN PUBLIC KEY-----\nMHYwEAYHK...AzztRs\n-----END PUBLIC KEY-----\n
+
+The `jwt_key` section lists all the keys that this CouchDB server trusts. You
```
wohali, line 415: Ideally this would link to the configuration portion of the documentation with the `:config:option` syntax.

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+should ensure that all nodes of your cluster have the same list.
+
+JWT tokens that do not include a `kid` claim will be validated against the
```
wohali, line 418: Double quotes

src/api/server/authn.rst
rnewson: Bad Request is the usual text for a 400, yes (c.f, https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#400)

src/api/server/authn.rst
rnewson: I will change this but I note single backticks is used everywhere else earlier in this file, including the 'proxy authentication' section I used as my base.

src/api/server/authn.rst
rnewson: I don't follow. your example is for an http request header, whereas this is a section of the config file.

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
 Note that you don't need to request :ref:`session <api/auth/session>`
 to be authenticated by this method if all required HTTP headers are provided.
+
+.. _api/auth/jwt:
+
+JWT Authentication
+====================
+
+.. note::
+    To use this authentication method make sure that the
```
flimzy, line 355: [grammar nit]

```suggestion
    To use this authentication method, make sure that the
```

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+    ``{chttpd_auth, jwt_authentication_handler}`` value is added to the
+    list of the active :config:option:`chttpd/authentication_handlers`:
+
+    .. code-block:: ini
+
+        [chttpd]
+        authentication_handlers = {chttpd_auth, cookie_authentication_handler}, {chttpd_auth, jwt_authentication_handler}, {chttpd_auth, default_authentication_handler}
+
+``JWT authentication`` enables CouchDB to use externally generated JWT tokens
```
flimzy, line 364: [grammar nit]

```suggestion
``JWT authentication`` enables CouchDB to use externally-generated JWT tokens
```

src/api/server/authn.rst
wohali: OK, I'll file another PR to fix all of that. Thanks for letting me know.

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+
+The ``alg`` claim is mandatory as it used to lookup the correct key for verifying the
+signature.
+
+The ``sub`` claim is mandatory and is used as the CouchDB user's name if the JWT token
+is valid.
+
+A private claim called ``_couchdb.roles`` is optional. If presented, it is a JSON
+array of strings, as is used as the CouchDB user's roles list if the JWT token
+is valid.
```
flimzy, line 396: Wording of the second sentence seems slightly awkward. Would this work?

```
A private claim called ``_couchdb.roles`` is optional. If present and valid, it is a JSON
array of strings, as is used as the CouchDB user's roles.
```

src/api/server/authn.rst
wohali: Actually, that is for a config section of `[couch_httpd_auth] x_auth_roles` with the text `X-Auth-CouchDB-Roles` as the hyperlink.

Here's a better example: 

```
When CouchDB sees a valid token in a subsequent request, it will
authenticate the user by this token without requesting the password again. By
default, cookies are valid for 10 minutes, but it's :config:option:`adjustable
<couch_httpd_auth/timeout>`. Also it's possible to make cookies
:config:option:`persistent <couch_httpd_auth/allow_persistent_cookies>`.
```

src/api/server/authn.rst
```diff
@@ -345,3 +345,124 @@ headers to CouchDB with related requests:
...
+The ``jwt_key`` section lists all the keys that this CouchDB server trusts. You
+should ensure that all nodes of your cluster have the same list.
+
+JWT tokens that do not include a ``kid`` claim will be validated against the
+``$alg:_default`` key.
+
+It is mandatory to specify the algorithm associated with every key for security
+reasons (notably presenting a HMAC-signed token using an RSA or EC public key
+that the server trusts:
+https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/).
```
flimzy, line 424: Suggestion:

```
For security reasons, it is mandatory to specify the algorithm associated with every key.
This prevents accepting a HMAC-signed token using an RSA or EC public key
that the server trusts:
https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/.

File: src/api/server/authn.rst (hunk 18 above)
Comment (rnewson) [reply src/api/server/authn.rst:355]: I don't agree that this is good grammar.

File: src/api/server/authn.rst (hunk 19 above)
Comment (rnewson) [reply src/api/server/authn.rst:364]: again, this is not conventionally hyphenated, I don't understand the suggestion that it should be.

File: src/api/server/authn.rst (hunk 20 above)
Comment (rnewson) [reply src/api/server/authn.rst:396]: should be "and is used as the" in the original comment. does that make more sense?

File: src/api/server/authn.rst (hunk 18 above)
Comment (flimzy) [reply src/api/server/authn.rst:355]: A subordinate clause before a main clause should be followed by a comma.  If you reverse the order, then no comma would be correct.

In either case, not a big deal.

File: src/api/server/authn.rst (hunk 19 above)
Comment (flimzy) [reply src/api/server/authn.rst:364]: It is correct to hyphenate here because it is an adjective phrase.  You would not hyphenate when it's used as an adverbial phrase. i.e. "The JWT is externally generated".

File: src/api/server/authn.rst [hunk 22]
```diff
@@ -345,3 +345,125 @@ headers to CouchDB with related requests:
...
+
+The ``alg`` claim is mandatory as it used to lookup the correct key for verifying the
+signature.
+
+The ``sub`` claim is mandatory and is used as the CouchDB user's name if the JWT token
+is valid.
+
+A private claim called ``_couchdb.roles`` is optional. If presented,
+it is a JSON array of strings, it is used as the CouchDB user's roles
+list as long as the JWT token is valid.
```
flimzy, line 397: Yes, makes much more sense now.  But I think you accidentally omitted "and" ?

src/api/server/authn.rst
```diff
@@ -345,3 +345,126 @@ headers to CouchDB with related requests:
...
+
+Two sections of config exist to configure JWT authentication;
+
+.. code-block:: ini
+
+    ;[jwt_auth]
+    ; List of claims to validate
+    ; required_claims =
+
+The :config:option:`required_claims <jwt_auth/required_claims>` config
```
wohali, line 385: This is great! But now you have to actually define this, otherwise the build fails because the link can't be constructed:

https://ci-couchdb.apache.org/blue/organizations/jenkins/jenkins-cm1%2FCouchDB-Documentation/detail/PR-524/12/pipeline

So I guess either revert this change, or add the config settings to the `src/config/auth.rst` or `misc.rst` file, following the patterns there.

I'd prefer the latter but will +1 the former.

PR #532, opened 2020-05-04
Correct the occurrence of _global_changes

mojito317:
## Overview

This PR intended to solve the missing document modifications from https://github.com/apache/couchdb/issues/2497. (Unfortunately, GitHub does not allow me to open the issue).

As per the 3.0.1 upgrade notes says
> The setup wizard no longer automatically creates the `_global_changes` database, as the majority of users do not need this functionality. This reduces overall CouchDB load.

If I am correct this means the default values of the `ensure_dbs_exist` property reduced from `["_users","_replicator","_global_changes"]` to `["_users","_replicator"]` at the `/_cluster_setup`. Moreover `/_db_updates` cannot be reached till the `_global_changes` database is not created beforehand.

## Checklist

### DIFF
--- a/src/api/server/common.rst
+++ b/src/api/server/common.rst
@@ -338,7 +338,7 @@
                      - :mimetype:`text/plain`
     :query array ensure_dbs_exist: List of system databases to ensure exist
         on the node/cluster. Defaults to
-        ``["_users","_replicator","_global_changes"]``.
+        ``["_users","_replicator"]``.
     :>header Content-Type: - :mimetype:`application/json`
                            - :mimetype:`text/plain; charset=utf-8`
     :>json string state: Current ``state`` of the node and/or cluster (see
@@ -437,7 +437,7 @@
         (add_node only)
     :<json array ensure_dbs_exist: List of system databases to ensure exist
         on the node/cluster. Defaults to
-        ``["_users","_replicator","_global_changes"]``.
+        ``["_users","_replicator"]``.
 
     *No example request/response included here. For a worked example, please
     see* :ref:`cluster/setup/api`.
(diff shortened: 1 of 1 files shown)


src/api/server/common.rst
```diff
@@ -453,7 +453,9 @@
 .. http:get:: /_db_updates
     :synopsis: Return the server changes of databases
 
-    Returns a list of all database events in the CouchDB instance.
+    Returns a list of all database events in the CouchDB instance. The
+    existence of the ``_global_changes`` database is required to use this
+    endpoint.
```
wohali, line 458: This is not true.

`/_db_updates` will function if `global_changes` is not present, but it will only show current data, not historical. That is, you cannot provide a `since` token and see historical values; it will act as if you always specify `?since=now`.

wohali:
@mojito317 Thank you for your first-time contribution to CouchDB! :field_hockey: :skier: :rugby_football:

There's a small change that needs to be made. If you can make this change, I'll be sure this gets into our next documentation update.

src/api/server/common.rst
mojito317: Interesting, because the endpoint does not show any data for me till I do not create the `_global_changes` database. 

```
curl --location --request GET 'localhost:5984/_db_updates' \
--header 'Authorization: ***'
```
Gives me a 404 with:
```json
{
    "error": "not_found",
    "reason": "Database does not exist."
}
```
@wohali what do you think, should I report this first?

wohali: You have to specify `continuous=true` as well, I believe, then try to create a DB while you're listening. Try that.

rnewson: hi,

It is true that you need the backing db ("_global_changes") for the _db_updates endpoint to work at all, regardless of feed type.

rnewson:
```cluster_system_dbs() ->
    ["_users", "_replicator"].``` from https://github.com/apache/couchdb/blob/master/src/setup/src/setup.erl#L77-L78

wohali: Thanks, I was mistaken, remembering how this worked before the backing store. 

Resolved.

mojito317:
Thanks a lot @rnewson and @wohali for the review!

Issue #517, opened 2020-04-06
Improvement to "3.1.2.2. Reduce and Rereduce Functions" argument description


toyssamurai:
Currently, under **3.1.2.2. Reduce and Rereduce Functions**, the description of the _keys_ argument is:

`Array of pairs of docid-key for related map function results. Always null if rereduce is running (has true value).`

## Expected Behavior
The description gives an impression that the value of the _keys_ argument would look like this:

`[ ["docid1", "key1"], ["docid2", "key2"] ... ]`

## Current Behavior
What I actually see is: 

`[ ["key1", "docid1"], ["key2", "docid2"] ... ]`

## Possible Solution
I think that it would be less confusing if the documentation says **key-docid** instead of **docid-key** or, better yet, gives an example of what the value of the _keys_ argument would look like.

wohali:
@toyssamurai Hey, thanks for this! This would be a very fast Pull Request, if you'd like to contribute the fix yourself. We'd certainly welcome the fix!

PR #533, opened 2020-05-04
Update ddocs.rst

sunvuz:
Improvement to "3.1.2.2. Reduce and Rereduce Functions" argument description #517

## Overview

Currently, under 3.1.2.2. Reduce and Rereduce Functions, the description of the keys argument is:
Array of pairs of docid-key for related map function results. Always null if rereduce is running (has true value).
The correct representation would be if the documentation says key-docid instead of docid-key

## Testing recommendations

Documentaion Correction. No testing required

## GitHub issue number

Closes https://github.com/apache/couchdb-documentation/issues/517

## Related Pull Requests

None

## Checklist

### DIFF
--- a/src/ddocs/ddocs.rst
+++ b/src/ddocs/ddocs.rst
@@ -97,7 +97,7 @@ Reduce and Rereduce Functions
 
 .. function:: redfun(keys, values[, rereduce])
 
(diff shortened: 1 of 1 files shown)


wohali:
Thank you @sunvuz ! :fireworks:

PR #548, opened 2020-05-22
limit and continous mode

ronag:
Document EOL for continous feed with limit.


wohali:
@ronag Thank you for your first contribution to CouchDB! :ringed_planet: :cat: :alien:

PR #549, opened 2020-05-23
Add missing word "key" in "What is a partition?"

Frederik-Baetens:
## Overview

I assume the author intended to write that sensor_id was the partition key, since it doesn't make sense to me how a field could be a partition.

## Checklist


janl:
thanks!

PR #557, opened 2020-06-03
Updated parameter details for _db_updates API method

vincnetas:
## Overview

Updated parameter details for _db_updates API method. As i'm not sure if this is expected behavior or defect, suggest to add this detail to parameter description.

## Checklist


src/api/server/common.rst
```diff
@@ -476,6 +476,7 @@
         keep the feed alive indefinitely. Default is ``60000``. May be ``true``
         to use default value.
     :query string since: Return only updates since the specified sequence ID.
+        If sequence ID is specified but does not exist, return all changes.
```
janl, line 479:
```suggestion
        If the sequence ID is specified but does not exist, all changes are returned.
```

PR #561, opened 2020-06-19
Allow drilldown for search to always be specified as list of lists

bessbd:
## Overview

To use multiple `drilldown` parameters users had to define
`drilldown` multiple times to be able supply them.

This caused interoperability issues as most languages require
defining query parameters and request bodies as associative
arrays, maps or dictionaries where the keys are unique.

This change enables defining `drilldown` as a list of lists so
that other languages can define multiple drilldown keys and values.

## Testing recommendations

`make check` passed for me locally

## Related Pull Requests

This is related to https://github.com/apache/couchdb/pull/2958

## Checklist

### DIFF
--- a/src/ddocs/search.rst
+++ b/src/ddocs/search.rst
@@ -814,7 +814,7 @@ Drilldown
 You can restrict results to documents with a dimension equal to the specified label.
 Restrict the results by adding ``drilldown=["dimension","label"]`` to a search query. You
 can include multiple ``drilldown`` parameters to restrict results along multiple
-dimensions.
+dimensions: ``drilldown=[["dimension1","label1"], ["dimension2","label2"]]``
 
 Using a ``drilldown`` parameter is similar to using ``key:value`` in the ``q`` parameter,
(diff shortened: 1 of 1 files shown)


src/ddocs/search.rst
```diff
@@ -814,7 +814,7 @@ Drilldown
 You can restrict results to documents with a dimension equal to the specified label.
 Restrict the results by adding ``drilldown=["dimension","label"]`` to a search query. You
 can include multiple ``drilldown`` parameters to restrict results along multiple
-dimensions.
+dimensions: ``drilldown=[["dimension1","label1"], ["dimension2","label2"]]``
```
rnewson, line 817: can you show an example of multiple parameters (like the sentence before indicates is allowed) and then clarify that this new example allows you to do this with a single drilldown parameter?

bessbd: Done in 5c02705 . Let me know what you think.

bessbd:
Thank you for the review and the merge, @rnewson !

PR #569, opened 2020-07-11
Improve explanations about scheduling windows

cluxter:
## Overview

This PR clarifies 2 points about scheduling window parameters in order to avoid confusion as in https://github.com/apache/couchdb/discussions/2997.


janl:
excellent, thanks!

PR #573, opened 2020-07-17
Update user doc password field details

natevw:
Adds a few missing updates/details regarding password fields within user doc, including:

1. `salt` field is used for **both** password schemes, not just ``simple``.
2. ``pbkdf2`` scheme also needs an `iterations` field
3. Added quick tip re. plaintext `password` field as well.

### DIFF
--- a/src/intro/security.rst
+++ b/src/intro/security.rst
@@ -285,18 +285,20 @@ several *mandatory* fields, that CouchDB needs for authentication:
 
 - **_id** (*string*): Document ID. Contains user's login with special prefix
   :ref:`org.couchdb.user`
-- **derived_key** (*string*): `PBKDF2`_ key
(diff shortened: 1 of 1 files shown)


natevw:
Full disclosure, I just noticed via https://github.com/apache/couchdb/blob/3e574519474662ddbc475e79c9c7f2f8a8fb4e25/src/couch/include/couch_js_functions.hrl#L67 that apparently CouchDB also supports a `bcrypt` scheme as well? So unless that's deprecated or something maybe those details should get added as well.

wohali:
We had to drop the bcrypt support, see https://github.com/apache/couchdb/pull/1457 for details. There are still traces in the code base.

janl:
@natevw otherwise looks good aside from the build fail ;)

wohali (approved):

PR #577, opened 2020-07-29
fix(api): Fix typo in ddoc/views.rst

fmoko:
## Overview

Fixes a small typo in `src/api/ddoc/views.rst`. Not sure if the wording as I've put it is correct, but given that there are probably multiple ICU drivers in existence, I'd wager it is.


flimzy:
LGTM! +1 and thanks!

janl:
FTR there is only one ICU driver :)

PR #585, opened 2020-09-01
Fix stale options

bessbd:
## Overview

The docs for `stale` in views don't match the current behavior of the software.
This PR fixes that issue by removing `false` from the available options from `stale` for views.

Mind that, the "default value" (`false`) for `stale` can no longer be expressed explicitly.

## Testing recommendations
```python3
print(requests.get(f"http://127.0.0.1:15984/{db_name}/_all_docs?stale=false", auth=('adm', 'pass')).json())
```
results in
`{'error': 'query_parse_error', 'reason': 'Invalid value for `stale`.'}`

## Related Pull Requests

This is a follow-up PR for https://github.com/apache/couchdb/pull/3120 .

## Checklist

### DIFF
--- a/src/maintenance/performance.rst
+++ b/src/maintenance/performance.rst
@@ -270,7 +270,7 @@ that will occur giving you a quick response and when views will be updated
 which will take a long time. (A 10 million document database took about 10
 minutes to load into CouchDB but about 4 hours to do view generation).
 
-In a cluster, "stable" requests are serviced by a fixed set of shards in order
+In a cluster, "stale" requests are serviced by a fixed set of shards in order
 to present users with consistent results between requests. This comes with an
 availability trade-off - the fixed set of shards might not be the most
 responsive / available within the cluster. If you don't need this kind of
--- a/src/maintenance/performance.rst
+++ b/src/maintenance/performance.rst
@@ -254,6 +254,8 @@ For example, something that takes 16 hex digits to represent can be done in
 Views
 =====
 
+.. _views/generation:
+
 Views Generation
 ----------------
 
--- a/src/maintenance/performance.rst
+++ b/src/maintenance/performance.rst
@@ -270,7 +270,7 @@ that will occur giving you a quick response and when views will be updated
 which will take a long time. (A 10 million document database took about 10
 minutes to load into CouchDB but about 4 hours to do view generation).
 
-In a cluster, "stale" requests are serviced by a fixed set of shards in order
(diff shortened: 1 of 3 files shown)


wohali:
@bessbd You also need to update this in `src/api/database/find.rst`.

If you want you can link to the Views Generation section in `maintenance/performance.rst` to help people understand the difference between `stale`, `stable` and `update`, where it's explained in detail.

bessbd:
> @bessbd You also need to update this in `src/api/database/find.rst`.
> 
> If you want you can link to the Views Generation section in `maintenance/performance.rst` to help people understand the difference between `stale`, `stable` and `update`, where it's explained in detail.

Thank you for the review, @wohali . I've pushed a commit 7a9094e to address your requests.

src/maintenance/performance.rst
```diff
@@ -268,7 +270,7 @@ that will occur giving you a quick response and when views will be updated
 which will take a long time. (A 10 million document database took about 10
 minutes to load into CouchDB but about 4 hours to do view generation).
 
-In a cluster, "stale" requests are serviced by a fixed set of shards in order
+In a cluster, "stable" requests are serviced by a fixed set of shards in order
```
wohali, line 273: Hm. This is a good change (not talking about `stale`), but down in line 279 it talks about `stale=ok` and `stale=update_after` again.

Similarly up in line 268 we talk about "stale" again.

If you want to fully expunge the concept of stale requests, we should do it in this entire section. I'd be +1 on that work but don't have time to help today.

Otherwise, I'd reverse the change and we can do that at a later date.

bessbd: > Hm. This is a good change (not talking about `stale`), but down in line 279 it talks about `stale=ok` and `stale=update_after` again.

Those lines refer to a deprecated parameter, while 0526b9b was fixing a misleading typo (the meaning by this is exactly the opposite of the behavior).

> 
> Similarly up in line 268 we talk about "stale" again.
> 
> If you want to fully expunge the concept of stale requests, we should do it in this entire section. I'd be +1 on that work but don't have time to help today.
> 
> Otherwise, I'd reverse the change and we can do that at a later date.

I only wanted to "leave the campground cleaner than I found it". To avoid scope creep in this PR, I've pushed 1fce16e to revert the fix for the misleading typo ( 0526b9b ).

rnewson (approved):

bessbd:
Thank you for the review and merge, @rnewson !

PR #587, opened 2020-09-04
Inline code block

samuela:
## Checklist


nickva (approved):
Thank you, @samuela

PR #594, opened 2020-10-02
Modify the second query of multi-query example.

asmand:
## Overview

The second query in the example only included a `skip` and a `limit` parameter. This requires a bit more thinking about the example.
The purpose of this PR is to make it very clear that the second query is a regular query too. 
It also tries to make the result set a bit more consistent.

## Testing recommendations

If the example given could be tested to verify that the output in the example is correct.

## GitHub issue number

n/a

## Related Pull Requests

n/a

## Checklist

n/a

### DIFF
--- a/src/api/database/bulk-api.rst
+++ b/src/api/database/bulk-api.rst
@@ -373,6 +373,8 @@ Sending multiple queries to a database
                 ]
             },
             {
+                "startkey": "pizza",
+                "endkey": "pizza\uFFFF",
                 "limit": 3,
                 "skip": 2
             }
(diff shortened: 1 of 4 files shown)


wohali:
@asmand I... guess? I'm not seeing the point of the change here. Was it not clear to you that the first query was `{"keys":["meatballs","spaghetti"]}` and the second query was `{"limit":3, "skip":2}`?

Why not keep the original request/response and insert a sentence above the example like:

> In the example below, two queries are run with one request. The first searches on the keys `meatballs` and `spaghetti`, while the second only uses a `limit` of 3 and a `skip` of 2.

This would prevent us having to recreate the example to get the sample output 🤣

PR #595, opened 2020-10-02
Modify multi query example main

asmand:
## Overview

The second query in the example only included a skip and a limit parameter. This requires a bit more thinking about the example.
The purpose of this PR is to make it very clear that the second query is a regular query too.
It also tries to make the result set a bit more consistent.

## Testing recommendations

If the example given could be tested to verify that the output in the example is correct.

## GitHub issue number

n/a

## Related Pull Requests

n/a

## Checklist

n/a


wohali:
Duplicate of #594

PR #596, opened 2020-10-05
cluster: add a blurb about changing name

sdahlbac:
## Overview

Rationale: this is something I would have wanted to see, to avoid having to change -name and reshard a bunch of databases.

As requested in https://github.com/apache/couchdb/pull/3188 I'll open this against the documentation instead.

However, I'm a bit on the fence if I would have looked at clustering docs or scrolled through vm.args, the discoverability aspect of it I mean.

❓ Should the PR be against `master` or against `main`?

## Checklist


wohali:
+1. Thanks so much for your first contribution to CouchDB!  🚴 🛹 ⛷️

PR #597, opened 2020-10-05
More inclusive terminology

sdahlbac:
## Overview

Switch to using more inclusive nomenclature

## Checklist

### DIFF
--- a/src/intro/consistency.rst
+++ b/src/intro/consistency.rst
@@ -268,7 +268,7 @@ consistency between multiple database servers. If a client makes a write
 operation on server `A`, how do we make sure that this is consistent with
 server `B`, or `C`, or `D`? For relational databases, this is a very complex
 problem with entire books devoted to its solution. You could use
-multi-master, single-master, partitioning, sharding, write-through caches,
+multi-primary, single-primary, partitioning, sharding, write-through caches,
 and all sorts of other complex techniques.
 
 Incremental Replication
--- a/src/config/auth.rst
+++ b/src/config/auth.rst
@@ -231,7 +231,7 @@ Authentication Configuration
             public_fields = first_name, last_name, contacts, url
 
         .. note::
-            Using the ``public_fields`` whitelist for user document properties
+            Using the ``public_fields`` allowlist for user document properties
             requires setting the :option:`couch_httpd_auth/users_db_public`
             option to ``true`` (the latter option has no other purpose)::
 
--- a/src/config/indexbuilds.rst
+++ b/src/config/indexbuilds.rst
@@ -49,7 +49,7 @@ following settings.
         database. If the difference is larger than the threshold defined here the
         background job will only be allowed to run in the main queue. Defaults to 1000.
 
-.. config:section:: ken.ignore :: Auto-Indexing Blacklist
+.. config:section:: ken.ignore :: Auto-Indexing Blocklist
 
 Entries in this configuration section can be used to tell the background indexer to skip
 over specific database shard files. The key must be the exact name of the shard with the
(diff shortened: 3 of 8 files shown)


janl (approved):

src/intro/consistency.rst
```diff
@@ -268,7 +268,7 @@ consistency between multiple database servers. If a client makes a write
 operation on server `A`, how do we make sure that this is consistent with
 server `B`, or `C`, or `D`? For relational databases, this is a very complex
 problem with entire books devoted to its solution. You could use
-multi-master, single-master, partitioning, sharding, write-through caches,
+multi-primary, single-primary, partitioning, sharding, write-through caches,
```
flimzy, line 271: I'm not sure if we came to an agreement on this term elsewhere, but "multi-primary" is a bit of an oxymoron, in a way that "multi-master" isn't.  Primary literally means "first", whereas "master" doesn't have that connotation.

If this is still open for discussion, perhaps we can consider "multi-leader" instead?

Or, my own 2¢, we can stick with "master", since this term isn't nearly as loaded as the "s" word, which CouchDB has never used.

garbados: Multi-leader :+1:

garbados (commented):
Glad to see this. Given the consternation about "multi-primary" I vote for "multi-leader" but I admit even that feels a bit clunky.

src/intro/consistency.rst
flimzy: > Glad to see this. Given the consternation about "multi-primary" I vote for "multi-leader" but I admit even that feels a bit clunky.

Agreed, "multi-leader" feels funny, too.  This is why I personally prefer the existing wording, since it's more precise, and AFAIK, I've never actually heard a compelling complaint against it. (Nobody's trying to get rid of "Masters Degree" or "Master of Cermonies", so I don't see why this technical term needs to suffer). I even recall a multi-year-old CouchDB issue on the topic, and the decision then was to stick with the current wording.  Granted, that was well before the recent surge of attention to this issue (which I _do_ welcome, for the record!)

sdahlbac: My thinking was that since we're moving away from master as a branch name, we should be consistent and move away from it elsewhere as well. 
I do see the point that @flimzy is making regarding the fact that there might be a stronger implication that there is only one primary.

Multi-leader is a bit funny, leader-leader replication possibly even more so.

Or, would 'main' work? Is multi-main and main-main replication be less awkward?

flimzy: Renaming the branch is easy, because there's no compelling reason to stick with `master`. There are many synonyms used in the wild that work just as well: `main`, `trunk`, and probably other less popular ones.  I don't think the same argument can be made for "Multi-master".  In this context, "master" is a technical term, that describes something very specific. Much like a "master key" or "master copy".

Perhaps alternatives to those can be found as well ("original key"?), but I'm not sure that translates to the database domain ("original-original sync"?)

I am, of course, in favor of inclusive verbiage. And while I don't _personally_ see a problem with a branch name of "master", it's so easy to change, that it's not really worth even discussing.  In this case, there's no clear alternative that doesn't somehow change the meaning (given current English usage--of course if the whole world switches to "primary" in this context, then it will take on that meaning, and this conversation, too, will be moot).

So TL;DR; I'm not suggesting any specific course of action. If we decide that "primary" (or "leader", or for that matter anything else ["rhubarb-rhubarb sync", anyone?]) is the term _we_ use, that's fine.  I guess my only point is that we should be aware, and intentional, about the change, as it may lead to confusion.

garbados: In the sense of *primus inter pares* or first-among-equals "primary" remains a valid verbiage, even if it is at first blush confusing.

But, I think I like "main" better. Consider a "main" and "alt", mirroring "primary" and "secondary". But there can be multiple mains -- since, with certain nuanced exceptions, they can answer for the same *main* dataset -- and one can engineer "alts" through replication patterns. But within a cluster, there are only mains. Is that too confusing?

sdahlbac: I wonder, would it make sense to split up this PR into two? 

The blacklist->blocklist, whitelist->allowlist and master->main **when talking about git branches** it seems no-one disagrees with. 

However, what (and even if) we should change master to is very much still up for debate.

sdahlbac:
This has now been split into and superceded by: #599 and #600  Closing.

PR #599, opened 2020-10-12
More inclusive terminology pt 1

sdahlbac:
## Overview

The original #597 included both bits that everyone seemed to agree with, and bits that need a bit more discussion before settling.
That pr has now been split into two, this one is the more straightforward one.

Switch to using more inclusive nomenclature

## Checklist


flimzy:
+1

flimzy (approved):

Issue #598, opened 2020-10-11
`POST /{db}/index` seems to desribe old behaviour.


b0wter:
I am referencing [this part](https://docs.couchdb.org/en/latest/api/database/find.html#db-index) of the documentation.

The _Example index creation using all available query parameters_ does not work for me. Posting it results in the following response:
```
{
    "error": "missing_required_key",
    "reason": "Missing required key: index"
}
```
I've played around with it a little and found that:

_The index object is a JSON array of field names following the sort syntax. Nested fields are also allowed, e.g. “person.name”._

no longer (?) holds true. It has (at least) two possibles properties: `fields` and `partial_filter_selector`. `partial_filter_selector` is listed as a query parameter but it shouldn't be.

An example using all available parameters looks like this:
```
{
  "index": {
    "partial_filter_selector": {
      "year": {
        "$gt": 2010
      },
      "limit": 10,
      "skip": 0
    },
    "fields": [
      "_id",
      "_rev",
      "year",
      "title"
    ]
  },
  "ddoc": "example-ddoc",
  "name": "example-index",
  "type": "json",
  "partitioned": false
}
```
I am running containerized version of CouchDb 3.1.1. I think the documentation is out of date and would update it according to my findings. In case I've not just misunderstood the documentation 😅

PR #601, opened 2020-10-13
Update for POST /{db}/_index

b0wter:
## Overview

I am opening this pr in reference to #598 
I've updated the following:

1. added a missing status code (404) 
2. removed `partial_filter_selector` from the query parameters (since it should be a property of the `index` parameter)
3. updated the `index` query parameter (it is no longer a list of strings but a json object)
4. changed the example that showcases the use of all query parameters

## Testing recommendations

Use any tool to send a http request to couchdb using the example given in the pr. I've tested it with Couch DB 3.1.1 and get the following response:
```
result": "exists",
    "id": "_design/example-ddoc",
    "name": "example-index"
}
```
(I've run multiple tests that's why the index already exists)

## Checklist

### DIFF
--- a/src/api/database/find.rst
+++ b/src/api/database/find.rst
@@ -987,17 +987,15 @@ built using MapReduce Views.
     :code 404: Database not found
     :code 500: Execution error
 
-**Index object format for JSON type indexes**
+    The `Index object` is a JSON object with the following fields:
 
-The index object is a JSON object with the following fields:
+    :json array fields: array of field names following the :ref:`sort
+       syntax <find/sort>`. Nested fields are also allowed, e.g. `"person.name"`.
+    :json json partial_filter_selector: A :ref:`selector <find/selectors>`
+       to apply to documents at indexing time, creating a
+       :ref:`partial index <find/partial_indexes>`. *Optional*
 
--  ``fields``: array of field names following the :ref:`sort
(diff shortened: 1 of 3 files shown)


src/api/database/find.rst
```diff
@@ -987,12 +984,18 @@ built using MapReduce Views.
...
+    :code 404: Database not found
     :code 500: Execution error
 
 **Index object format for JSON type indexes**
 
-The index object is a JSON array of field names following the :ref:`sort
-syntax <find/sort>`. Nested fields are also allowed, e.g. `"person.name"`.
+The index object is a JSON object with the following fields:
+
+-  ``fields``: array of field names following the :ref:`sort
```
wohali, line 994: This section should be formatted similar to this: 

https://github.com/apache/couchdb-documentation/blame/main/src/api/server/common.rst#L644-L658

Use the same `:json ...` type structure so that all of the fields render correctly.

wohali (changes requested):
Minor formatting changes - overall this looks good.

src/api/database/find.rst
b0wter: I've updated the PR. Here is a screenshot of what the new part looks like:

image

Let me know if this is not the way it's supposed to be :)

wohali (approved):

wohali:
Thanks @b0wter !

PR #610, opened 2020-12-15
Revert "Suggested edit to make clear what a quorum actually represent…

rnewson:
This reverts commit 4eabf7978a61d4e4432cfd4f5e647179e39f3661.

This notion of different quorums for doc read vs doc write is not true.


davisp (approved):
+1

wohali:
Also +1

PR #618, opened 2021-01-27
Update changes.rst

tmdoit:
Fix typo

## Checklist

### DIFF
--- a/src/api/database/changes.rst
+++ b/src/api/database/changes.rst
@@ -24,7 +24,7 @@
     resource. Only the most recent change for a given document is guaranteed to
     be provided, for example if a document has had fields added, and then
     deleted, an API client checking for changes will not necessarily receive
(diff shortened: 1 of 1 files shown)


wohali:
Hey @tmdoit , this is actually correct in the original.

The idea is that if you add a document to CouchDB, then change it a couple of times, and _then_ replicate to another DB, the intermediate _document_ versions won't be reflected - you only get the latest version of the document.

This may be added fields, but it may also be changed or deleted ones. The point is that intermediate document revisions are discarded.

I'm going to close this provisionally, but if you feel you can take the explanation above and work that into a new PR that clarifies the text for you in that way, I'm happy to entertain it.

tmdoit:
Hi @wohali, it make sense now, no change is required, just conscious reading... thanks for your time and have a good day!

PR #620, opened 2021-02-03
Fix Unicode Collation Algorithm link

tmdoit:
## Checklist


wohali:
Thanks!

PR #621, opened 2021-02-05
Fix anchor to retrieving deleted documents

nono:


flimzy:
This doesn't appear to fix a link--the link looks the same to me. It seems to just move it. Was that the intention? If so, maybe the description can be updated to explain why this is helpful?

nono:
In the ` DELETE /{db}/{docid}` section, there is a link called "Retrieving deleted documents". When clicking on it, we are sent to a section that is not about retrieving deleted documents, but about retrieving attachments. I have moved the anchor, so that the link will go to the right section (the one called "Retrieving Deleted Documents").

flimzy:
Ah, makes sense! Thanks for the explanation.

flimzy:
+1

PR #639, opened 2021-03-29
Improve mango docs

### LINKED ISSUE #350: Improve use_index documentation
Comes from apache/couchdb#1740:

>I am having some trouble with use_index and partial indexes in CouchDB 2.2.0. This part of the _find documentation is misleading:
>
> > Technically, we don’t need to include the filter on the "status" field in the query selector - the partial index ensures this is always true - but including it makes the intent of the selector clearer and will make it easier to take advantage of future improvements to query planning (e.g. automatic selection of partial indexes).
>
> The documentation should add that an index with fields will never be used unless the selector includes all the fields indexed. ( `"fields": []` works around this problem ). This is different than a typical SQL database that would use an index for queries against any prefix of the list of indexed fields.
>
> I was also confused by using the index's `name` in `use_index`.  Fauxton should put `ddoc` in its example index definitions to make it very clear that the index `name` cannot be used in `use_index` unless it is accompanied by `ddoc`.
>
> Given a design doc like this one,
>
>`{"ddoc": "_design/obsolete-created", "type": "json", "name": "foo-json-index", "def": {"fields": [{"meta.state": "asc"}, {"meta.created": "asc"}], "partial_filter_selector": {"$and": [{"type": "form"}, {"meta.state": "obsolete"}]}}}`
>
```
In [172]: db.explain({"use_index":"obsolete-created", "selector":{"meta.state":"obsolete"}})
Out[172]: 
{u'dbname': u'db',
 u'fields': u'all_fields',
 u'index': {u'ddoc': None,
  u'def': {u'fields': [{u'_id': u'asc'}]},
  u'name': u'_all_docs',
  u'type': u'special'},
 u'limit': 25,
 u'mrargs': {u'conflicts': u'undefined',
  u'direction': u'fwd',
  u'end_key': u'<MAX>',
  u'include_docs': True,
  u'reduce': False,
  u'stable': False,
  u'start_key': None,
  u'update': True,
  u'view_type': u'map'},
 u'opts': {u'bookmark': u'nil',
  u'conflicts': False,
  u'execution_stats': False,
  u'fields': u'all_fields',
  u'limit': 25,
  u'r': [49],
  u'skip': 0,
  u'sort': {},
  u'stable': False,
  u'stale': False,
  u'update': True,
  u'use_index': [u'obsolete-created']},
 u'selector': {u'meta.state': {u'$eq': u'obsolete'}},
 u'skip': 0}
```
>
```
In [175]: db.explain({"use_index":"obsolete-created", "selector":{"meta.state":"obsolete", "meta.created": "foobar"}})
Out[175]: 
{u'dbname': u'db',
 u'fields': u'all_fields',
 u'index': {u'ddoc': u'_design/obsolete-created',
  u'def': {u'fields': [{u'meta.state': u'asc'}, {u'meta.created': u'asc'}],
   u'partial_filter_selector': {u'meta.state': {u'$eq': u'obsolete'}}},
  u'name': u'obsolete-created',
  u'type': u'json'},
 u'limit': 25,
 u'mrargs': {u'conflicts': u'undefined',
  u'direction': u'fwd',
  u'end_key': [u'obsolete', u'foobar', u'<MAX>'],
  u'include_docs': True,
  u'reduce': False,
  u'stable': False,
  u'start_key': [u'obsolete', u'foobar'],
  u'update': True,
  u'view_type': u'map'},
 u'opts': {u'bookmark': u'nil',
  u'conflicts': False,
  u'execution_stats': False,
  u'fields': u'all_fields',
  u'limit': 25,
  u'r': [49],
  u'skip': 0,
  u'sort': {},
  u'stable': False,
  u'stale': False,
  u'update': True,
  u'use_index': [u'obsolete-created']},
 u'selector': {u'$and': [{u'meta.created': {u'$eq': u'foobar'}},
   {u'meta.state': {u'$eq': u'obsolete'}}]},
 u'skip': 0}
```

Joan writes:

> @dholth Some of these rules have changed with checkins since 2.2.0 was released: https://github.com/apache/couchdb/commit/8e28fd2af5732d1b7e2866977860695fcbd676e8
> 
> I agree the documentation could be improved, and am moving this issue to our couchdb-documentation repository.
> 
> We'd love a pull request if you are interested in helping out! :wink:

### LINKED ISSUE #380: Document Mango inequality selector(s) behavior more accurately
In the Mango selector documentation, the [Condition Operators section](https://docs.couchdb.org/en/2.3.0/api/database/find.html#condition-operators) has a table for each of the operators. I think some clarification of these is warranted.

For example, the purpose of `$ne` is currently listed as:

> The field is not equal to the argument.

However, it appears that this definition is incomplete. The way it behaves in practice is more like:

> The field **exists, and** is not equal to the argument.

See for example this Stack Overflow question on [CouchDB/PouchDB selector for deleted docs](https://stackoverflow.com/questions/43571099/couchdb-pouchdb-selector-for-deleted-docs) where the OP was surprised when `_deleted: {$ne: true}` did not match any documents, even though `_deleted: {$exists: false}` does.

The documentation should make the behavior clearer, since naïvely I too would have assumed `$ne` works like e.g. `!=` or `!==` in JavaScript, where `doc._deleted !== true` *would* match documents where the `'deleted_'` field is missing.

I suspect this currently-implicit "field exists, and…" may be true of all the inequality operators, but only tested for `$ne` specifically.

### LINKED ISSUE #381: Document strictness of Mango comparison operators
The Mango [Condition Operators documentation](https://docs.couchdb.org/en/2.3.0/api/database/find.html#condition-operators) does not specify how equality matches are performed with respect to argument type. Is any coercion (between e.g. string and number, or number and boolean) ever performed?

Testing this in practice with a document `{_id:"test", value: 1}`:

* the selector `value: {$eq: 1}` matches
* the selector `value: {$eq: "1"}` does not match

So the current behavior appears strict, which may not be assumed by developers coming from languages like JS/PHP that can include looser comparison operations.

### LINKED ISSUE #391: Are _find API docs missing an entry for conflicts key?
I've successfully followed the docs to use a Mango query to scan for conflicted documents:

https://docs.couchdb.org/en/latest/replication/conflicts.html#finding-conflicted-documents-with-mango

However I've noticed that the example query includes a `conflicts` boolean key, which isn't mentioned in the `_find` API docs:

https://docs.couchdb.org/en/latest/api/database/find.html#db-find

Is this an omission? If so, I'm happy submit a patch for the docs, but what would be a good description for the key?

wohali:
* Improve use_index documentation, closes #350
* Clarify condition operators, closes #380 
* Clarify comparison ops, closes #381
* Include conflicts key, closes #391

### DIFF
--- a/src/api/database/find.rst
+++ b/src/api/database/find.rst
@@ -654,10 +654,12 @@ specified field contains a value that is equal to the supplied argument.
     the specified field must exist, and is not equal to the value of the
     argument.
 
-The basic equality and inequality operators common to most programming languages
-are supported. In addition, some 'meta' condition operators are available. Some
-condition operators accept any valid JSON content as the argument.
-Other condition operators require the argument to be in a specific JSON format.
+The basic equality and inequality operators common to most programming
+languages are supported. Strict type matching is used.
+
+In addition, some 'meta' condition operators are available. Some condition
+operators accept any valid JSON content as the argument.  Other condition
+operators require the argument to be in a specific JSON format.
 
 +---------------+-------------+------------+-----------------------------------+
 | Operator type | Operator    | Argument   | Purpose                           |
--- a/src/api/database/find.rst
+++ b/src/api/database/find.rst
@@ -44,6 +44,9 @@
     :<json string|array use_index: Instruct a query to use a specific index.
         Specified either as ``"<design_document>"`` or
         ``["<design_document>", "<index_name>"]``. *Optional*
+    :<json boolean conflicts: Include conflicted documents if ``true``.
+        Intended use is to easily find conflicted documents, without an
+        index or view. Default is ``false``. *Optional*
     :<json number r: Read quorum needed for the result. This defaults to 1, in
         which case the document found in the index is returned. If set to a
         higher value, each document is read from at least that many replicas
--- a/src/api/database/find.rst
+++ b/src/api/database/find.rst
@@ -648,6 +648,12 @@ Condition operators are specific to a field, and are used to evaluate the value
 stored in that field. For instance, the basic `$eq` operator matches when the
 specified field contains a value that is equal to the supplied argument.
 
+.. note::
+    For a condition operator to function correctly, the field **must exist**
+    in the document for the selector to match. As an example, ``$ne`` means
+    the specified field must exist, and is not equal to the value of the
+    argument.
+
 The basic equality and inequality operators common to most programming languages
 are supported. In addition, some 'meta' condition operators are available. Some
 condition operators accept any valid JSON content as the argument.
--- a/src/api/database/find.rst
+++ b/src/api/database/find.rst
@@ -1153,7 +1153,7 @@ it easier to take advantage of future improvements to query planning
 (e.g. automatic selection of partial indexes).
 
 .. note::
-    An index with fields will never be used, unless the selector includes
+    An index with fields is only used, when the selector includes
     all of the fields indexed. For instance, if an index contains ``["a". "b"]``
     but the selector only requires field ``["a"]`` to exist in the matching
     documents, the index would not be valid for the query. All indexes,
--- a/src/api/database/find.rst
+++ b/src/api/database/find.rst
@@ -1141,6 +1141,14 @@ but including it makes the intent of the selector clearer and will make
 it easier to take advantage of future improvements to query planning
 (e.g. automatic selection of partial indexes).
 
+.. note::
+    An index with fields will never be used, unless the selector includes
+    all of the fields indexed. For instance, if an index contains ``["a". "b"]``
+    but the selector only requires field ``["a"]`` to exist in the matching
+    documents, the index would not be valid for the query. All indexes,
+    however, can be treated as if they include the special fields ``_id`` and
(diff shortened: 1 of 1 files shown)


src/api/database/find.rst
```diff
@@ -1141,6 +1152,14 @@ but including it makes the intent of the selector clearer and will make
 it easier to take advantage of future improvements to query planning
 (e.g. automatic selection of partial indexes).
 
+.. note::
+    An index with fields will never be used, unless the selector includes
```
bessbd, line 1156: Do you think double-inverting this sentence would make it easier to understand?
```suggestion
    An index with fields is only used, when the selector includes
```

src/api/database/find.rst
```diff
@@ -1141,6 +1152,14 @@ but including it makes the intent of the selector clearer and will make
...
+    all of the fields indexed. For instance, if an index contains ``["a". "b"]``
```
bessbd, line 1157: From this example, I understand that
| Selector | Index | Index is used |
|----------|-------|---------------|
| a        | a, b  | no            |
| a, b     | a     | yes           |

Is that correct?

bessbd (approved):
LGTM, +1

PR #640, opened 2021-03-29
Warn against use of new_edits, closes #390

### LINKED ISSUE #390: behaviour of new_edits=false is not well documented
I often must deeply explain to other users, how new_edits works, how to use it (or why to not use it). The new_edits flag is mentioned along with _bulk_docs without any description or link to an appropriate section

## Possible Solution

* Describe that new_edits=false has purpose for replication only. There should be link to the section 2.4.2.5.2 where replication is documented
* Describe format of the document, when new_edits is set to false. It should mention that document must include the key "_revisions" otherwise bad things may happen.
 * There is mention about the key "_revisions" in the section "2.4.2.5.2. Upload Batch of Changed Documents", but the format of the key is not documented.  There should be a link to the section "2.18. Returned CouchDB Document with Revision Info". 
* It may be a good idea to describe, how _revisions is used to connect history of the document updates and when the conflict may happen. It may help to user to understand purpose of this field and why its presence is mandatory.

wohali:
Closes #390

### DIFF
--- a/src/api/document/common.rst
+++ b/src/api/document/common.rst
@@ -199,7 +199,8 @@
       and ``false``. If ``false``, a well-formed ``_rev`` must be included in
       the document. ``new_edits=false`` is used by the replicator to insert
       documents into the target database even if that leads to the creation
-      of conflicts. *Optional*
+      of conflicts. *Optional*, **The ``false`` value is intended for use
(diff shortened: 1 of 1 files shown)


nickva (approved):
+1

PR #660, opened 2021-06-02
Update documentation with couch_prometheus endpoint API

tonysun83:
## Overview

In the upcoming 3.2 release, we added supported for prometheus endpoints:
https://github.com/apache/couchdb/issues/3377
https://github.com/apache/couchdb/pull/3416
https://github.com/apache/couchdb/pull/3506

This updates our documentation to reflect this new addition.

## GitHub issue number

https://github.com/apache/couchdb/issues/3377
## Checklist


src/api/server/common.rst
```diff
@@ -1586,6 +1586,75 @@ node, you can use:
...
+============================
+``/_node/{node-name}/_prometheus``
+=============================
+
+.. http:get:: /_node/{node-name}/_prometheus
+    :synopsis: Returns server statistics in prometheus format
+
+    The ``_prometheus`` resource returns a text/plain response that consolidates our
+    `_stats`, `active_task`, and `_system` endpoints. The format is determined by
+    https://prometheus.io/docs/introduction/overview/. Currently the format version 2.0.
```
flimzy, line 1598: I feel like this last sentence is missing a word?  "Currently the format version is 2.0" perhaps?

Also, we can probably remove the word "Currently", as I think that's implicit.

src/api/server/common.rst
```diff
@@ -1586,6 +1586,75 @@ node, you can use:
...
+        couchdb_erlang_reductions_total 46527253
+        # TYPE couchdb_erlang_processes gauge
+        couchdb_erlang_processes 528
+        # TYPE couchdb_erlang_process_limit gauge
+        couchdb_erlang_process_limit 262144
+        couchdb_active_task{type="replication", source="mailbox", target="http://mailsrv:5984/mailbox <http://mailsrv:5984/mailbox>", docs_count = "docs_read"} 4524
+        ouchdb_active_task{type="replication", source="mailbox", target="http://mailsrv:5984/mailbox <http://mailsrv:5984/mailbox>", docs_count = "docs_written"} 4524
+        couchdb_active_task{type="replication", source="mailbox", target="http://mailsrv:5984/mailbox <http://mailsrv:5984/mailbox>", docs_count = "missing_revisions_found"} 4524
+
+If an additional port config option is specified, then a client can call this api using a
```
flimzy, line 1647:
```suggestion
If an additional port config option is specified, then a client can call this API using
```

src/api/server/common.rst
```diff
@@ -1586,6 +1586,75 @@ node, you can use:
...
+that port which does not require authentication. This option is false(OFF) by default.
```
flimzy, line 1648:
```suggestion
that port which does not require authentication. This option is ``false`` (OFF) by default.
```

src/api/server/common.rst
```diff
@@ -1586,6 +1586,75 @@ node, you can use:
...
+When the option true(ON), the default ports for a 3 node cluster are 17986, 27986, 37986.
```
flimzy, line 1649:
```suggestion
When the option is ``true`` (ON), the default ports for a 3 node cluster are ``17986``, ``27986``, and ``37986``.
```

src/config/misc.rst
```diff
@@ -269,3 +269,37 @@ Configuration of Database Purge
...
+        Sets whether or not to create a separate, non-authenticated port (default is false)::
+
+            [prometheus]
+            additional_port = true
+
+    .. config:option:: bind_address
+
+        .. versionadded:: 3.2
+
+        The ip address to bind::
```
flimzy, line 293:
```suggestion
        The IP address to bind::
```

src/config/misc.rst
```diff
@@ -269,3 +269,37 @@ Configuration of Database Purge
...
+
+            [prometheus]
+            bind_address = 127.0.0.1
+
+    .. config:option:: port
+
+        .. versionadded:: 3.2
+
+        The port which clients can query prometheus endpoint data without authentication
```
flimzy, line 302:
```suggestion
        The port on which clients can query prometheus endpoint data without authentication
```

flimzy (commented):
A few minor formatting and grammar suggestions.

tonysun83:
@flimzy I've updated the documentation with your suggestions. Let me know what you think.

src/api/server/common.rst
```diff
@@ -1586,6 +1586,76 @@ node, you can use:
...
 
+==================================
+``/_node/{node-name}/_prometheus``
+==================================
+
+.. http:get:: /_node/{node-name}/_prometheus
+    :synopsis: Returns server statistics in prometheus format
+
+    The ``_prometheus`` resource returns a text/plain response that consolidates our
+    `_stats`, `active_task`, and `_system` endpoints. The format is determined by
```
flimzy, line 1597:
```suggestion
    ``_stats``, ``active_task``, and ``_system`` endpoints. The format is determined by
```

Maybe link to the docs about these endpoints, too?

src/api/server/common.rst
```diff
@@ -1586,6 +1586,76 @@ node, you can use:
...
+    https://prometheus.io/docs/introduction/overview/. The format version is 2.0.
```
flimzy, line 1598:
```suggestion
    https://prometheus.io/docs/introduction/overview/. The format version is 2.0.
```

Make this URL a link, too?

flimzy:
> @flimzy I've updated the documentation with your suggestions. Let me know what you think.

Looks good! I think the `ip` => `IP` capitalization was missed. And on a second look, I found a couple other suggestions.

src/api/server/common.rst
wohali: +1 to Flimzy's comment here

src/api/server/common.rst
wohali: +1

src/api/server/common.rst
```diff
@@ -1586,6 +1586,76 @@ node, you can use:
...
+        couchdb_erlang_context_switches_total 358276
+        # TYPE couchdb_erlang_reductions_total counter
+        couchdb_erlang_reductions_total 46527253
+        # TYPE couchdb_erlang_processes gauge
+        couchdb_erlang_processes 528
+        # TYPE couchdb_erlang_process_limit gauge
+        couchdb_erlang_process_limit 262144
+        couchdb_active_task{type="replication", source="mailbox", target="http://mailsrv:5984/mailbox <http://mailsrv:5984/mailbox>", docs_count = "docs_read"} 4524
+        ouchdb_active_task{type="replication", source="mailbox", target="http://mailsrv:5984/mailbox <http://mailsrv:5984/mailbox>", docs_count = "docs_written"} 4524
+        couchdb_active_task{type="replication", source="mailbox", target="http://mailsrv:5984/mailbox <http://mailsrv:5984/mailbox>", docs_count = "missing_revisions_found"} 4524
```
wohali, line 1645: Is there any explanation as to why this subset of the metrics are available, vs. the full metrics we can see at `_system` or `_stats`?

tonysun83: @wohali it was a rather long set of metrics and I cut it short not to take up a lot of space. I can include the entire thing so that it doesn't confuse users.

wohali: Oh no need! Just put an ellipsis in and state that it contains the full content of `_stats` and `_system` maybe? Or if there are intentional omissions, list that too. Even a bulleted list of what to expect would be sufficient.

tonysun83: added an ellipsis (I think I did it correctly) in there. This was a good catch. I had mistakenly added `_active_tasks` into the documentation because originally I wanted the PR to include `_active_tasks`. But upon inspecting the entire list output I realized I didn't have it. I went back to the PR and I remembered that `_active_tasks` is a little different given how it's actually something we accumulate at the top of chttpd and not per node. That's why I excluded it for this first version. The next improvement would be to dig deeper grab tasks per node.

tonysun83:
@flimzy @wohali made the changes you both recommended. will be away for a bit so feel free to squash and merge if there are no more changes you'd like to see

Issue #661, opened 2021-06-03
Adding an information about max_partition_limit in the Partition Database introduction section


Shin-Aska:
During the introduction of partitions, it indicates that the general recommendation for partition size is 10gb.

Perhaps it would also be a good idea to introduce max_partition_size option to the readers early on to clearly tell them that the default partition size is indeed 10gb and can be changed or disabled by modifying the said option.

PR #662, opened 2021-06-03
Added a note regarding max_partition_size in index.rst

Shin-Aska:
Added a small note in the What is a partition section introducing the reader to the ``max_partition_size`` option. This small note should make it clear that the default limit (and not just a recommendation) for partitions is 10gb but can be changed or disabled accordingly as mentioned in issue #661


nickva (approved):
+1

Thank you for your contribution

PR #667, opened 2021-06-28
Fix escaping of quotes for restructured text

henokgetachew:
## Overview

This fixes erlang command used to connect the two nodes in order to check connectivity.

## Checklist


nickva:
That is an erlang code block, don't think we should escape single quotes for atoms?

wohali:
@nickva is right, this looks correct to me.

Snap 2021-07-22, 12_14_21

PR #700, opened 2021-11-18
Fix statement around collation

mojito317:
## Overview

Meanwhile, the documentation says `collation` should be set to `raw` in the root of the design document's `options` object, actually, it can be set under the view object where `options` will be a sibling of the `map` and `reduce` fields:

```json
{
  "_id":"_design/doc",
  "views": {
     "myview1": {
         "map": "function(...) {}",
         "options": {"collation" : "raw"}
     }
  }
}
```

This is misleading, because design doc root also can have an `options` object, but the children of that object can be  `include_design`, `local_seq`, and `partitioned`.

## Testing recommendations

No

## GitHub issue number

not a significant change

## Related Pull Requests

no

## Checklist

I tested with `make check`.

### DIFF
--- a/src/api/ddoc/views.rst
+++ b/src/api/ddoc/views.rst
@@ -638,9 +638,9 @@ By default CouchDB uses an `ICU`_ driver for sorting view results. It's possible
 use binary collation instead for faster view builds where Unicode collation is
 not important.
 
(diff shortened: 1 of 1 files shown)


nickva (approved):
+1

Thank you for your contribution!

PR #705, opened 2022-01-31
skip/limit won't work properly when _design docs exists

jiahuili430:
## Related Pull Requests
https://github.com/apache/couchdb/pull/3832

## Checklist

### DIFF
--- a/src/api/server/common.rst
+++ b/src/api/server/common.rst
@@ -219,6 +219,12 @@
            "locations"
         ]
 
+.. note::
+    The "`skip`" and "`limit`" parameters of GET `_all_dbs` will not work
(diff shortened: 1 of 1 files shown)


nickva (commented):
We had added an internal _design doc there at some point from custodian. It's now removed but users (operators) may add their own and this should document the corner case.

We should fix it at some point but until then is good to have.

Since we're updating the API notes, let's add two separate notes, one for `GET _all_dbs` and one for `GET _dbs_info`

jiahuili430:
Changed, thanks for the review. @nickva

nickva (approved):
+1

PR #710, opened 2022-04-01
Fix typo (missing literal block mark)

cailloumajor:
## Overview

Before this fix, the configuration example of https://docs.couchdb.org/en/3.2.0/config/couchdb.html#couchdb/default_security is not rendered correctly:

image

Well-rendered example:

image


flimzy (approved):
+1

PR #724, opened 2022-05-05
Update reverse proxy config for nginx

### LINKED ISSUE #675: Reverse Proxying CouchDB in a subdirectory with nginx example does not handle URL encoded path fragments
## Expected Behavior

Using the rewrite setup described in https://docs.couchdb.org/en/stable/best-practices/reverse-proxies.html#reverse-proxying-couchdb-in-a-subdirectory-with-nginx I would expect nginx to pass my path parameters 1:1.

E.g. I request:

```
GET /couchdb/_users/org.couchdb.user%foo.bar%2Btest8%40example.com?conflicts=true
```

and the request that CouchDB sees is:

```
GET /_users/org.couchdb.user%foo.bar%2Btest8%40example.com?conflicts=true
```

## Current Behavior

Using the documented rewrite rules, nginx decodes the path parameters, which means CouchDB sees this instead:

```
GET /_users/org.couchdb.user:foo.bar+test8@example.com?conflicts=true
```

which makes CouchDB return 404 (I would assume this is because of the `+`).

## Possible Solution

Instead, the following rewrite rules can be used:

```
location /couchdb {
    rewrite ^ $request_uri;
    rewrite ^/couchdb/(.*) /$1 break;
    proxy_pass http://localhost:5984$uri;
    proxy_redirect off;
    proxy_buffering off;
    proxy_set_header Host $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
}
```
See: https://serverfault.com/questions/459369/disabling-url-decoding-in-nginx-proxy

## Steps to Reproduce (for bugs)

1. Create user doc whose id contains a `+`
2. Use the reverse proxy setup in a subdirectory as described in the docs
3. GET the user doc
4. 404 response

big-r81:
Fix #675

### DIFF
--- a/src/best-practices/reverse-proxies.rst
+++ b/src/best-practices/reverse-proxies.rst
@@ -104,8 +104,9 @@ as ``http://domain.com/couchdb/db1/doc1`` are proxied to
 .. code-block:: text
 
     location /couchdb {
-        rewrite /couchdb/(.*) /$1 break;
-        proxy_pass http://localhost:5984;
+        rewrite ^ $request_uri;
+        rewrite ^/couchdb/(.*) /$1 break;
+        proxy_pass http://localhost:5984$uri;
         proxy_redirect off;
         proxy_buffering off;
         proxy_set_header Host $host;


nickva (approved):
+1

PR #725, opened 2022-05-06
Remove statement about single replica scale

garethbowen:
With one replica you can have more than one node so long as you have multiple shards.


garethbowen:
I'm not 100% confident that my analysis is correct though I have tested it and it seems to be true. If I'm right, then I'm not sure exactly what the original author was trying to say so I've just removed it.

nickva (approved):
That makes sense. Good observation. A q=8,n=1 might spread replicas across a few nodes.

PR #736, opened 2022-05-23
Clarifies encoding of peruser database names

DougReeder:
## Overview

Existing documentation was ambiguous.  This should make it unambiguous.


big-r81 (approved):
+1

Thank you for your contribution.

PR #742, opened 2022-06-10
Notes _find performance difference between custom indexes and _all_docs

DougReeder:
## Overview

The docs currently don't contain much guidance on the performance of _find, which can lead new users down the wrong path.


nickva (approved):
+1

Thank you for your contribution

PR #745, opened 2022-07-05
fix typo.

alishir:
## Overview

Fix typo.


big-r81 (approved):
Good catch! Thank you.

PR #747, opened 2022-07-11
Documents how to add and remove nodes if the Cluster Setup Wizard isn't enough.

DougReeder:
## Overview

Documents how to manually fix cluster management when the automated tool isn't able to do the right thing.

Also changes the name of a link to the name of the setting it references.


nickva (approved):
Thank you for your contribution, @DougReeder!

+1

PR #748, opened 2022-07-15
Migrate to rebar3

jiahuili430:
## Checklist


nickva (approved):
+1 seems like a no-op change mostly and would work with existing logic

